PR #70, opened 2015-01-04
Fixed ignoring of http redirection limit

markvandertol:
First: You've created a great library! Thank you for that.

A small issue I noted when I looked at the code. When the status code is `HTTP_MOVED_PERM` or `HTTP_SEE_OTHER`, it ignores the request count. This is because the `&&` has a higher operator precedence and therefore get evaluated before the `||` operators.


davidgyoung:
Good catch, thanks!

Issue #71, opened 2015-01-04
Class load warnings on Android 4.4 and lower


markvandertol:
I compiled the `master` branch and tried it on my Android 4.4 phone, but it shows a lot of warnings on LogCat. It still works, but useless warnings are of course undesired. 

```
01-04 21:03:34.111 E/dalvikvm﹕ Could not find class 'android.bluetooth.le.ScanCallback', referenced from method org.altbeacon.beacon.service.CycledLeScanner.finishScanCycle
01-04 21:03:34.111 W/dalvikvm﹕ VFY: unable to resolve check-cast 32 (Landroid/bluetooth/le/ScanCallback;) in Lorg/altbeacon/beacon/service/CycledLeScanner;
01-04 21:03:34.121 W/dalvikvm﹕ Unable to resolve superclass of Lorg/altbeacon/beacon/service/CycledLeScanner$4; (32)
01-04 21:03:34.121 W/dalvikvm﹕ Link of class 'Lorg/altbeacon/beacon/service/CycledLeScanner$4;' failed
01-04 21:03:34.121 E/dalvikvm﹕ Could not find class 'org.altbeacon.beacon.service.CycledLeScanner$4', referenced from method org.altbeacon.beacon.service.CycledLeScanner.getNewLeScanCallback
01-04 21:03:34.121 W/dalvikvm﹕ VFY: unable to resolve new-instance 452 (Lorg/altbeacon/beacon/service/CycledLeScanner$4;) in Lorg/altbeacon/beacon/service/CycledLeScanner;
01-04 21:03:34.121 W/dalvikvm﹕ VFY: unable to resolve virtual method 60: Landroid/bluetooth/BluetoothAdapter;.getBluetoothLeScanner ()Landroid/bluetooth/le/BluetoothLeScanner;
01-04 21:03:34.121 E/dalvikvm﹕ Could not find class 'android.bluetooth.le.ScanSettings$Builder', referenced from method org.altbeacon.beacon.service.CycledLeScanner.scanLeDevice
01-04 21:03:34.121 W/dalvikvm﹕ VFY: unable to resolve new-instance 35 (Landroid/bluetooth/le/ScanSettings$Builder;) in Lorg/altbeacon/beacon/service/CycledLeScanner;
[...]

PR #76, opened 2015-01-11
Split old implementation and new implementation for CycledLeScanner

markvandertol:
As discussed in ticket #71

I couldn't test it properly on a L device, since I don't have one yet. However, I tried not to make changes to the logic, but only move things around, so everything should work.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -23,14 +23,10 @@
  */
 package org.altbeacon.beacon.service;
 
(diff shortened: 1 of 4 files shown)


davidgyoung:
Thanks, @markvandertol, in general this looks good.  I am going to merge these changes onto the android-5-scanning-optimizations branch, so I can properly test.

davidgyoung:
I have merged this into a new branch here:

https://github.com/AltBeacon/android-beacon-library/tree/android-5-scanning-optimize-split

I need to do some real testing to see if everything is working correctly

PR #77, opened 2015-01-11
Don't export services in AndroidManifest

markvandertol:
Because the services are only used from this library, they don't have to be exported in the manifest. Prevents potential security issues.

Specifying `exported=false` on services without an `intent-filter` element isn't necessary as `false` is then the default in that case, but this makes it more explicit when people read the manifest.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -16,14 +16,15 @@
         </receiver>
 
         <service android:enabled="true"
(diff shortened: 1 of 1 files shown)


davidgyoung:
Thanks for the catch, @markvandertol.  This is a holdover from the original design of the service when it was intended to be a single service used by multiple applications.  You are correct about the reasons it should no longer be exported.

Issue #81, opened 2015-01-13
Release build defaulting to snapshot with wrong file name


davidgyoung:
after doing a `./gradlew release` with the tag 2.1-beta2, it looks like the version string is incorrectly getting suffixed with SNAPSHOT because I see:

    Building with version=2.1-beta2-SNAPSHOT

And then the files that get generated don't have the version at all in the suffix.  The contents of build/outputs/aar shows:

    android-beacon-library-release.aar	
    android-beacon-library-release.tar.gz

@pieces029, any thoughts here?

pieces029:
The snapshot gets appended if you don't pass gradle the -Prelease flag. I
can see why this would be confusing for a task and a flag to have the same
name.

pieces029:
We can change this up, it's just the way I do it since it's harder to accidentally publish something or check in the wrong version number this way. Thoughts?

davidgyoung:
Understood about the extra protection of the -Prelease flag.  That's fine by me.

Any thoughts on why the filenames don't become 

    android-beacon-library-2.1-beta2.aar  
    android-beacon-library-2.1-beta2.tar.gz

when I do:  `./gradlew release -Prelease` ?  Instead they are:

    android-beacon-library-release.aar	
    android-beacon-library-release.tar.gz

pieces029:
I think that's just how the assembleRelease task creates the aar's. I'll take a look and see if I can find anything about the gz files.

pieces029:
The tar file comes from bundle-eclipse script, basing it off of the name of the release aar. Did it not always use to be like this? I think the only compilation tasks I added where to create jar files of sources and javadocs.

davidgyoung:
Looking at this a bit more... yeah, this is not the way it used to be.  With the current setup, when you do a release build it always creates a local aar filename without a version number like this: /build/output/aar/android-beacon-library-release.aar .

When you upload it to bintray, it gets renamed to the version number, e.g. android-beacon-library-2.1-beta2.aar 

Since it gets renamed in the upload, that's all for the aar.  Except that the .tar.gz file (which is used for Eclipse and converted from the aar using the bundle-eclipse script) also ends up without a version number in the filename.  This is a problem because this file cannot be distributed on bintray and has to be manually uploaded to a release on Github so people can download it the old fashioned way.  So it has to be manually renamed to android-beacon-library-2.1-beta2.tar.gz first.

[...]

PR #85, opened 2015-01-17
Add tar task

pieces029:
- Created a task to bundle the eclipse tar.gz, fixing #81 
- Added tar.gz to be included in bintray. 
  - Unsure if this will effect how things are published to jcenter and central. I have only had archive type formats in there own repo in bintray.


davidgyoung:
Nice to get rid of a shell script!  I will test tomorrow.

davidgyoung:
Looks good.  I tested it an the generated .tar.gz file works as a library project in Eclipse.

One addition I would like to make is to *also* make a copy of the .aar file with the same version identifier in the filename, so I can easily upload this as a GitHub release, too.

What I get now is:

android-beacon-library						
android-beacon-library-debug.aar
android-beacon-library-2.1-beta3-7-g50c374f-SNAPSHOT.tar.gz
android-beacon-library-release.aar

What I would like is:

android-beacon-library						
android-beacon-library-debug.aar
android-beacon-library-2.1-beta3-7-g50c374f-SNAPSHOT.tar.gz
android-beacon-library-2.1-beta3-7-g50c374f-SNAPSHOT.aar
android-beacon-library-release.aar

I tried playing with the way you set it up to add this, but I can't seem to figure out how the filename name gets assigned to the .tar.gz file

pieces029:
It's gradle wizardry. The tar task by default grabs that name from somewhere. I'll add another copy task so you get the output you want. I think I might move this stuff to a separate gradle file to try and keep the main one clean.

pieces029:
Updated

Issue #64, opened 2014-12-29
Use Logging Interface


pieces029:
Currently all logs are logged out no matter what build type I am doing.

I would prefer something like slf4j, or even just a simple interface I could set to provide my own logging.

davidgyoung:
Fixed in #87

PR #87, opened 2015-01-17
Update logging

pieces029:
Updated the logging to have all log messages wrapped, allowing users to not have any logging, or specify a custom logger.

Sorry if this PR is too noisy. I can revert some of the changes if you would like.


pieces029:
#64

src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -138,27 +126,184 @@ public static void setDebug(boolean debug) {
...
+    /**
+     * Set the logger to be used with in the AltBeacon library. This will receive all log message
+     * the library outputs.
+     *
+     * @param logger The logger to be used.,
+     * @throws NullPointerException if logger is null.
+     * @see org.altbeacon.beacon.Loggers
+     * @since 2.1
+     */
+    public static void setsLogger(Logger logger) {
```
davidgyoung: This should probably be named setLogger not setsLogger -- I'm guessing this was code generation that blindly ignored the s prefix on the static member variable.

src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -138,27 +126,184 @@ public static void setDebug(boolean debug) {
...
      */
-    public static final long DEFAULT_BACKGROUND_BETWEEN_SCAN_PERIOD = 5*60*1000;
+    public static final long DEFAULT_BACKGROUND_BETWEEN_SCAN_PERIOD = 5 * 60 * 1000;
 
     private long foregroundScanPeriod = DEFAULT_FOREGROUND_SCAN_PERIOD;
     private long foregroundBetweenScanPeriod = DEFAULT_FOREGROUND_BETWEEN_SCAN_PERIOD;
     private long backgroundScanPeriod = DEFAULT_BACKGROUND_SCAN_PERIOD;
     private long backgroundBetweenScanPeriod = DEFAULT_BACKGROUND_BETWEEN_SCAN_PERIOD;
 
+    private static Logger sLogger = Loggers.empty();
```
davidgyoung: I would still like the default logger to be Loggers.androidLogger().  Most Android developers will be looking at LogCat during their development process, and it is important for them to see errors and warnings produced by the library.  If the default is Loggers.empty() then people won't know when things go wrong unless they have taken action to set up an alternate logger.

src/main/java/org/altbeacon/beacon/BeaconManager.java
pieces029: Yeah, I think it was a blind refactor. I'll update this. I realized too, that this is a breaking change, but since this is a whole minor change update I'm guessing it's okay.

davidgyoung:
Cool.  In general I think this looks good.

I do have a concern not on these specific changes but on reworking the logging in general.  I worry that there is a lot of processing time taken up preparing strings for debug logging (string concatenation and the like), even though debug logging is turned off by default.  This can especially be a problem for debug lines inside low-level scanning.  The two common approaches to this are to (a) remove those log lines or (b) wrap them in an if statement like `if(BeaconManager.debug)` around where the log string is constructed.  I don't want to remove the log lines, because I run into issues all the time where capturing detailed debug logs is critical to finding end-users' problems. (e.g. #84)  This leaves option (b) which makes the code a bit uglier.  Any thoughts here?

I know this concern is probably a big beyond the scope of what you were trying to achieve with this pull request, but if we are going to make a global logging change, then this seems like the right time to fix this issue as well.

src/main/java/org/altbeacon/beacon/BeaconManager.java
pieces029: I would argue that in a release library there shouldn't be logging by default, and with the right amount of documentation I think people would know what to do for development. 

We could do `sLogger = BuildConfig.DEBUG ? Loggers.androidLogger() : Loggers.empty()` by default, but I think that would probably be more confusing. I'll update it to `androidLogger` implementation if you would like. I think I'll update for the loggers to lazy load as well, might as well have a smaller footprint if we can.

I did notice it seems like some of the docs in the website are out of date, I'll do a pass through and update what I can.

pieces029:
I didn't notice any excessive string concatenation. I'll take a look, but string literals should be placed in the string pool and shouldn't really cause performance issues. Do you have any performance measurements on this?

davidgyoung:
I don't have numbers because it is hard to measure the impact without having a way to globally remove all debug statements for testing.  The case I am worried about is when there are many beacons visible and the app is in the background.  Even if debug is turned off, about 40 strings get generated for each detected beacon.  A single beacon is typically detected 10 times per second, so if 10 beacons are visible, this is 10*10*40 = 4000 String objects created every second, and since several have some string concatenation operations this can be more.

I wish Java had a pre-processor like C so a release build could remove all the debug log lines while a development/debug build could leave them in.  But as it stands the only way I know to address this is by wrapping each one with an if statement.

src/main/java/org/altbeacon/beacon/BeaconManager.java
davidgyoung: I do prefer that  the androidLogger be the default.  Even if folks aren't running with BuildConfig.DEBUG, I'd like warnings and errors to show up in LogCat even if people have not configured a logger.

src/main/java/org/altbeacon/beacon/BeaconManager.java
davidgyoung: Why do you say this is a breaking change?

pieces029: Since I removed the public setDebug method. Anyone that was using that will have to change to the setLogger method.

pieces029:
I get what you are saying. This sounds like a perfect use case for slf4j, but I get not wanting to have a transitive dependency. It will be a little bit more involved but I can make something similar that won't do the concatenation unless it's actually being written out somewhere.

src/main/java/org/altbeacon/beacon/BeaconManager.java
davidgyoung: I see.  i still think it makes sense to allow control the log level, which is what setDebug is designed to do.  I say leave that method in, and retain the default behavior, which is android logging with debug level disabled.

markvandertol:
You could create for every log method a overload that also accepts parameters as vararg. The allocation of a small array is much less work than the allocation of new String (or even specific implementations for 1, 2, etc. arguments). The Logger implementation is then responsible for calling [String.format](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#format(java.lang.String,%20java.lang.Object...)) when the log level is enabled. A log statement doesn't have that many parameters, so providing some overloads probably won't be much work.

I think it would be better when the static log methods would be in a separate class instead of part of BeaconManager. They aren't really related to the other functionality of the BeaconManager.

pieces029:
@markvandertol Yeah that's similar to what I was planning. I kept the logging in BeaconManager since it seemed like the main place for configuring this library.

davidgyoung:
A reasonable middle ground might be to have convenience methods on BeaconManager to configure logging (set the level and set the logger), but have those just call merods on a dedicated class, which is what hosts the .d, .I, .e methods.

pieces029:
Haven't had time to update this yet. Will get around to it soon.

davidgyoung:
Thanks, @pieces029 .

pieces029:
@davidgyoung Updated the who logging framework. 

How do you feel about `BluetoothCrashResolver` still wrapping the debug logs? I would rather just deprecated those methods and have them not doing anything, but wanted your input.

Also made this so there are no breaking changes.

src/main/java/org/altbeacon/beacon/logging/LogManager.java
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Manager for logging in the Altbeacon library. The default is a
+ * {@link org.altbeacon.beacon.logging.Loggers#debug()}.
+ *
+ * @author Andrew Reitz
+ * @since 2.2
+ */
+public final class LogManager {
+    private static Logger sLogger = Loggers.debug();
```
davidgyoung: By default, it appears this will use the debug logger, which will output everything (debug, info, warning, error) to LogCat, correct?  I think the default should be that it only outputs only warning and error messages.

davidgyoung:
This is looking really good, @pieces029.  See my one comment about default behavior.  

On the issue of the logging in the BluetoothCrashResolver, I don't have any strong opinions.  Its logging had been set up slightly differently because the class was copied from a standalone repo which did not have the BeaconManager.  I think it is fine the way you have it now.

src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -46,81 +48,87 @@
...
- *	      			Log.i(TAG, "The first beacon I see is about "+beacons.iterator().next().getDistance()+" meters away.");		
- *     			}
- *        	}
- *  		});
- *  		
+ * 	      			Log.i(TAG, "The first beacon I see is about "+beacons.iterator().next().getDistance()+" meters away.");
+ *                }
+ *            }
+ *        });
+ * <p/>
```
markvandertol: This <code>&lt;p/&gt;</code> breaks the layout in Android Studio. In a <code>&lt;pre&gt;</code> block a paragraph tag isn't needed to get line breaks.

pieces029:
Updated to use a default warning logger that will log warnings and above.
Because this is the default implementation, I deprecated the debug flags in the BluetoothCrashResolver and those methods no longer to anything.

davidgyoung:
Looks good.  Glad to see you brought in Mockito, too.  I have been meaning to do that myself.

PR #92, opened 2015-01-18
Also execute other AsyncTask in ThreadPool

markvandertol:
#89 changed one AsyncTask, but two others in the code. Those were left unchanged.

* Perform downloading of distance config file in AsyncTask also in a ThreadPool, so the download doesn't block other AsyncTasks.
* Removed AsyncTask for BluetoothCrashResolver since doInBackground was called directly, so it was actually called synchronously,

### DIFF
--- a/src/main/java/org/altbeacon/beacon/distance/ModelSpecificDistanceCalculator.java
+++ b/src/main/java/org/altbeacon/beacon/distance/ModelSpecificDistanceCalculator.java
@@ -3,6 +3,7 @@ package org.altbeacon.beacon.distance;
 import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.pm.PackageManager;
+import android.os.AsyncTask;
 import android.os.Build;
 import android.util.Log;
 
@@ -20,7 +21,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Properties;
 
 /**
  * Obtains a <code>DistanceCalculator</code> appropriate for a specific Android model.  Each model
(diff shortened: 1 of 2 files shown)


davidgyoung:
Thanks, this looks right, @markvandertol  FYI, the only reason the BluetoothCrashResolver had the (unused) AsyncTask is because the code was copied from a [standalone variant](https://github.com/RadiusNetworks/bluetooth-crash-resolver) of the BluetoothCrashResolver where it is necessary.

I think it is great to remove that unused code from the BluetoothCrashResolver, but I want to make sure it is properly tested first.  It's actually pretty hard to execute that code.  You need to have a device on 4.3-4.4.4, and put it near a Bluetooth LE transmitter with a randomized mac address until it has too many bluetooth devices in its internal table, which triggers a Bluetooth crash and the crash resolver, forcing it to start the discovery process and then cancel the discovery process.  That's pretty time consuming to test!  

If you think you can do this testing and attach a LogCat showing it all works properly, then we can merge that part of the pull request.  Otherwise, I think it is probably not worth the risk to include the changes in BluetoothCrashResolver.java when merging this.

Thoughts?

pieces029:
This probably sounds crazy, but any chance we could create something with unit tests (on the jvm or on android) to simulate this? The actual test setup seems like way to much work.

markvandertol:
AsyncTask wasn't used in the right way. Calling doInBackground() directly is something you shouldn't do, since then it won't offload the work to another thread. So this code should be exactly the same. But I understand you want to be sure it works correctly.

Writing a test that verifies how BluetoothCrashResolver interacts with the other classes shouldn't be too hard write. There is a chance that the test isn't close enough to a real world usage, so you cannot verify with certainty that it prevents the crashes. But I think the number of interactions with BluetoothAdapter are limited, so writing a test sounds as an effective way here to verify that it still works correctly. Agreed or do you think it is still too risky?

davidgyoung:
It looks like I can test this end-to-end by manually calling forceFlush() on BluetoothCrashResolver

davidgyoung:
OK, using the technique I described above, I verified there is no difference in the behavior before or after this change of the cancellation of the bluetooth discovery process:

```
W/BluetoothCrashResolver(16759): Recovery attempt started
W/BluetoothCrashResolver(16759): BluetoothAdapter.ACTION_DISCOVERY_STARTED never received.  Recovery may fail.
W/BluetoothCrashResolver(16759): Recovery attempt finished
```

So this looks good to me, I will merge as soon as the continuous integration server is happy again

Issue #93, opened 2015-01-19
Update Bintray Plugin to 1.0


pieces029:
It looks like a lot of the confusing rest client tasks can be removed and the same tasks can be accomplished now with the Bintray plugin's DSL.

davidgyoung:
I believe this is merged and complete.

PR #94, opened 2015-01-19
Updated to use bintray plugin 1.0

pieces029:
@davidgyoung You'll have to test this out with the next release to make sure it's working as expected, but moving the tar.gz upload into the bintray task fixes the snapshot publishing issue.

 #93


davidgyoung:
ok, let's give it a shot

PR #98, opened 2015-01-25
Changed Region to become immutable.

markvandertol:
I've changed the Region class to become immutable of instantiation. Therefore it is no longer needed to have a copy constructor or for the object to be cloneable.

Also fixed the constructor that supported custom length Identifier lists and let equals take into account the identifiers.

Maybe for the backwards compatibility it is better to bring back the public clone method, maybe in deprecated form? What do you think?

### DIFF
--- a/src/main/java/org/altbeacon/beacon/Region.java
+++ b/src/main/java/org/altbeacon/beacon/Region.java
@@ -25,7 +25,6 @@ package org.altbeacon.beacon;
 
 import android.os.Parcel;
 import android.os.Parcelable;
-import android.util.Log;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/org/altbeacon/beacon/Region.java
```diff
@@ -161,19 +160,27 @@ public boolean matchesBeacon(Beacon beacon) {
...
-	
-	public String toString() {
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Region region = (Region) o;
+
+        if (!mIdentifiers.equals(region.mIdentifiers)) return false;
```
davidgyoung: I think this line is a problem.  The use of the uniqueId string to identify the region is intended to let you stop ranging or monitoring the region (or update the region being monitored/ranged) by passing in a second region with the same uniqueId string.  iOS CoreLocation uses the same basic mechanism.  If you construct a new region object with a different identifier set and want the APIs to range/monitor the modified region instead, it is supposed to *replace* that region in the ranging/monitoring list.  This wouldn't work anymore with this line added.

davidgyoung:
I doubt many people are using the public clone method, but yes, I do think it is better to have it be there and be deprecated until the next release with a breaking api change.

markvandertol:
Thank you for having a look. I've put back the original equals and added back a public clone method.

davidgyoung:
Looks good, @markvandertol.

Issue #97, opened 2015-01-25
Improve design of Identifier class


markvandertol:
I was looking at the Identifier class. I have some ideas to make the implementations of that class simpler. It also has a couple of bugs.

Most important improvements:

* It is confusing to have a single parse method that accepts all three forms of input, namely decimal, hexadecimal and uuid. I think clarity would be improved if there would be a specific one for each type and one generic that accepts a byte array.
* Make instances of this class immutable, the value no longer changes after it has been constructed, and therefore safe to share between threads. This is because other threads are unable to change the content of the instances after creation. This makes it also makes it safe to share this object with other objects without cloning, because it cannot change after it has been created. This makes it much easier to reason about concurrency and prevents bugs such as race conditions.

Some other smaller improvements:

* Strings aren’t the most efficient way to use as a normalized way to store the data. Byte arrays would be much simpler.
* Private no-param constructor isn’t needed when there is already another constructor.
* compareTo method follows the characteristics of the Java Comparable interface. Adding that interface makes it easier to use that method in other use cases.

Bugs I found:

* The Identifier constructor accepts null, but then most methods will throw null-pointer exceptions. I propose just to reject nulls in the constructor.
* Regular expressions in the class Identifier contain the following character classes: [0-9A-F-a-f]. The hyphen between the ‘F’ and the ‘a’ cause the ‘-’ to be an acceptable character. It will accept 2F234454-CF---4A0F-ADF2-F4911BA9FFA6 as an uuid.
* Only decimals are accepted with max length of 2 bytes. BeaconParser considers everything up to 4 bytes decimals (unsigned). Identifier can then return two different exceptions: Integer.parseInt() when it overflows or the check that the number is less than 65536. When a byte array is used to back the value, then the length of the identifier no longer matters.

If you also think that this is an improvement I could send in a pull request that refactors this class. The public API of this class stays compatible with the current version. What do you think?

davidgyoung:
Good ideas, @markvandertol.  The only one that worries me a bit is disallowing nulls.  I want to be careful this does not break backward compatibility.  An example of your proposal here might be useful.

Totally agree about the byte arrays, too. I have been meaning to do this for awhile, but have been put off by the effort to redactor the accompanying serialization in the Beacon class.

markvandertol:
[...]

PR #103, opened 2015-02-01
Changed Identifier class to work with a byte array instead of a String

markvandertol:
Made the class immutable, so it is no longer necessary to copy the Identifier before returning it.

As discussed in #97 

A breaking change I noticed was that parsing a value in decimal format no longer retains leading zeros, as SBeaconTest shows. I don't think this matters much. This doesn't affect the hexadecimal format.


davidgyoung:
Looks good.  I agree that the S-Beacon breaking change is minor and acceptable.  I highly doubt it will affect anybody.

Issue #116, opened 2015-02-12
ConcurrentModificationException


furkanvarol:
Recently I started to have this exception. Sometimes it crashes the app and sometimes don't. I will try to investigate the problem.

```
E/AndroidRuntime﹕ FATAL EXCEPTION: main
    java.util.ConcurrentModificationException
            at java.util.HashMap$HashIterator.nextEntry(HashMap.java:806)
            at java.util.HashMap$KeyIterator.next(HashMap.java:833)
            at org.altbeacon.beacon.BeaconManager$1.onServiceConnected(BeaconManager.java:574)
            at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1114)
            at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1131)
            at android.os.Handler.handleCallback(Handler.java:733)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:136)
            at android.app.ActivityThread.main(ActivityThread.java:5579)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:515)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1268)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1084)
            at dalvik.system.NativeStart.main(Native Method)
```

mattwritescode:
@furkanvarol this is already fixed correct? Or is this somehow separate from https://github.com/AltBeacon/android-beacon-library/pull/117?

furkanvarol:
I mixed my work email in #117 so I push again in #118. If u got the particular error that wrote in first comment than yes, it is fixed.

You can check it out, it was a single line fix actually. I just removed Map's put method in a  foreach (BeaconManager's service connection handler).

PR #117, opened 2015-02-12
fixed ConcurrentModificationException on BeaconManager

furkanvarol:
I investigated the problem I told before (#116). It seems like 94520a8813fa6abd6582a72290c44b459f66fb0d caused problem. I have fixed the problem (actually reverted code). Since, ConsumerInfo is not an immutable class, putting the reference of it after changing isConnected to true is unnecessary.

So, it could be in this way:
```
for (BeaconConsumer consumer : consumers.keySet()) {
    Boolean alreadyConnected = consumers.get(consumer).isConnected;
    if (!alreadyConnected) {
        consumer.onBeaconServiceConnect();
        ConsumerInfo consumerInfo = consumers.get(consumer);
        consumerInfo.isConnected = true;
    }
}
```

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


furkanvarol:
I think I should re-open pr since my work email got mixed as user.email. Sorry for that

PR #118, opened 2015-02-12
fixed ConcurrentModificationException on BeaconManager

furkanvarol:
I investigated the problem I told before (#116). It seems like 94520a8 caused problem. I have fixed the problem (actually reverted code). Since, ConsumerInfo is not an immutable class, putting the same reference to the map after changing isConnected to true is unnecessary.

So, it could be in this way as well:

```
for (BeaconConsumer consumer : consumers.keySet()) {
    Boolean alreadyConnected = consumers.get(consumer).isConnected;
    if (!alreadyConnected) {
        consumer.onBeaconServiceConnect();
        ConsumerInfo consumerInfo = consumers.get(consumer);
        consumerInfo.isConnected = true;
    }
}
```

BTW, sorry for the last PR.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -28,7 +28,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


davidgyoung:
I do not understand why this change makes a difference.  It just uses an `Iterator`  instead of a `for( : )` loop.  Why does this avoid a `ConcurrentModificationException`?  Seems like what needs to happen is making a copy of the Set before looping through them.

furkanvarol:
Beacuse in the foreach loop you can not change the collection and in last line in foreach tries to add an item (actually set) to the collection. Like I said before, I looked at the commits and you had this problem before and fixed the way I did it. That is why I used Iterator instead of other possible solutions.

davidgyoung:
I see.  I think it is actually clearer to remove the `consumers.put(consumer, consumerInfo);` line and and keep the for loop as you suggested as an alternative in your comment on the PR.

furkanvarol:
Yes I think that way is better and will be a lot faster. If you want I can create another PR for that.

davidgyoung:
Yeah, or you can just commit to this branch and comment on the change as part of this PR.  Same purpose.

furkanvarol:
I added a commit to change fix to a clearer way as you requested.

PR #123, opened 2015-02-16
fix concurrency exception when multiple consumer bound on start

furkanvarol:
My old fix was not working with multiple consumer is bound at same time and also some of them is unbound before getting onBeaconServiceConnect event. I investigated code deeper this time, I saw you put synchronized blocks but it seems like those blocks does not working properly. No matter what I did except changing HashMap to ConcurrentHashMap, my use case was problematic. Actually, I don't know why ConcurrentModificationException occurs since synchronized blocks should already cover this problem.

BTW, I know my use case a little bit complicated since I need to bind multiple consumers and unbind them quickly still this can be a major issue for someone else too.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -23,21 +23,6 @@
  */
 package org.altbeacon.beacon;
 
-import java.security.acl.LastOwnerException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import org.altbeacon.beacon.logging.LogManager;
-import org.altbeacon.beacon.logging.Loggers;
-import org.altbeacon.beacon.service.BeaconService;
-import org.altbeacon.beacon.simulator.BeaconSimulator;
-import org.altbeacon.beacon.service.StartRMData;
-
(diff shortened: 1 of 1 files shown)


davidgyoung:
The code looks good.  I am a little concerned that I have not been able to reproduce, and neither of us seem to understand why the synchronized blocks did not cover the problem.

@furkanvarol, how much testing have you done with your app that quickly binds/unbinds since you made this change?  How confident are you that it resolves the issue?

furkanvarol:
Yes I know it is weird that synchronized blocks not preventing concurrent exception but I deeply checked and tried thing at the end none of them solved the exception except `CourrentHashMap`.

About my case, I was doing a late initialization of `RegionBootstrap` and it were problematic for me (I don't know why) but creating a `RegionBootstrap` and disabling it immediately were solving the problem (this trick required once and at the beginning of the app). Yes, this is also weird but that was the only solution that make my app worked. I didn't create an issue for this because I thought that maybe it is my fault but for concurrent exception, I think this must be applied nevertheless none of the user will do what I do exactly but this proved that concurrent exception may occur.

Also, I have checked why a map with synchronized block causes `ConcurrentModificationException` and I couldn't find anything similar. So, I can not say that this pr must be merged but `CourrentHashMap` provides cool features for synchronization which makes this pr should be merged :). And, yes with this fix my problems with `ConcurrentModificationException` for this case solved for sure.

davidgyoung:
This appears to work properly based on my tests.

PR #126, opened 2015-02-17
Fix links to sample projects.

vladris:
Looks like Android Studio reference project is in main and Eclipse in a branch but these links were the other way around.


mattwritescode:
Good catch @vladris! Thanks for making the change!

Issue #114, opened 2015-02-09
Beacon Tracking Cache


IgorZaus:
I wonder, if it was a good idea to implement a tracking cache. If scanning period and beacon frequency is about the same (say, one second), it happens quite often, that a beacon disappears for one scan period. In real life this is rather unlikely, so I suggest to change 
RangeState.finalizeBeacons() 
to something like that:

if (rangedBeacon.getTrackingAge() > RangedBeacon.MAX_TRACKING_AGE)
   rangedBeacon.setTracked(false);

and in RangeBeacon.addMeasurement() 
[...]
lastTracked = System.currentTimeMillis();
[...]

and implement

public long getTrackingAge() {
        return System.currentTimeMillis() - lastTracked;
}

I don't know if there are better methods for that. Currently, I implemented that behaviour on the client side, when I get the callback from the BeaconService and it works as exected and gives better results in terms of stability.

davidgyoung:
Yes, I often implement this on my client code, too.  I agree it may be a good idea to build this in to the library.  Because some apps may need the current behavior, any tracking cache should have a configurable interval that can be set to 0.

Perhaps this can be combined with #113 as it touches the same code.  

@IgorZaus , any interested in putting these changes together and sending as a pull request?

IgorZaus:
Yes, give me a couple of days to prepare some sources!

IgorZaus:
Created a pull request - this is my very first contribution to Gitub - hope, I did everything ok....

davidgyoung:
Thanks, @IgorZaus.  Looks like you did the procedure correctly, thanks!  I need to review the changes and do some testing.  I will be out of the country for the next week, but I will try this out when I am back.

davidgyoung:
Implemented as described.

ersin-ertan:
Could some documentation be provided with this functionality. I'm not understanding:

`BeaconManager.setUseTrackingCache();`
`beaconManager.setMaxTrackingAge();`

- what it does
- for what situation you would want to modify these values. 
- what the implications of not using this is
- how do the two methods work together

davidgyoung:
The methods are intended to be used separately.  If  `setUseTrackingCache(true)` is called, then beacons will be tracked forever by Ranging APIs, and will never disappear from the callbacks.  

By default this is set to false, and they expire and are no longer returned in the ranging callback if they have not been seen in the number of milliseconds indicated by `setMaxTrackingAge(...)`, which defaults to 5000.

Issue #113, opened 2015-02-09
Alternative RSSI Measurement


IgorZaus:
Hi, 

I suggest an alternative way of measure the RSSI value. Instead of measuring a plain moving average I'd suggest to calculate an auto regressive moving average. The advantages would include less memory consumption, faster, smoother and better control over the smoothness.

An very naive implementation would be as follows:

/*
 * This RangedBeacon calculates its rssi on base of an auto regressive moving average (ARMA)
 * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
 * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
 * Note: a smoother average needs longer to "settle down"
 */
public class ARMARangedBeacon extends RangedBeacon {

    private static final String TAG = "ARMARangedBeacon";
    //initially set to min value
    private int armaMeasurement = -100;

    public ARMARangedBeacon(Beacon beacon) {
        super(beacon);
    }

	public void addMeasurement(Integer rssi) {
        LogManager.d(TAG, "adding rssi: %s", rssi);
        setTracked(true);
        armaMeasurement = Double.valueOf(armaMeasurement - 0.1*(armaMeasurement - rssi)).intValue();
        LogManager.d(TAG, "armaMeasurement: %s", armaMeasurement);
	}

	public boolean noMeasurementsAvailable() {
		return false;
	}

	protected double calculateRunningAverage() {
		return armaMeasurement;

	}

Of course instantiation in RangeState.addBeacon() should be configurable and not just:

 mRangedBeacons.put(beacon, new ARMARangedBeacon(beacon));

davidgyoung:
I agree a more flexible and configurable filtering of RSSI is a good idea.  Perhaps it could be designed to have a RssiFilter interface, and the active RssiFilter could be set on the BeaconManager.  The RangedBeacon could use whatever RssiFilter is configured.

A couple of different filters could be built in -- RunningAverageRssiFilter (default) and AmraRssiFilter, each with configrable parameters.

davidgyoung:
This has been implemented in #113 and #114

samskiter:
Hey @IgorZaus would you mind explaining the difference between your ARMA filter and an IIR. I can't connect the dots between ArmaRssiFilter and the wikipedia explanation of an ARMA model: https://en.wikipedia.org/wiki/Autoregressive%E2%80%93moving-average_model

[...]

PR #128, opened 2015-02-18
Changes for #113 and #114

IgorZaus:
RSSI Filter Implementation via Interfaces and Tracking Cache

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -23,21 +23,6 @@
  */
 package org.altbeacon.beacon;
 
-import java.security.acl.LastOwnerException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import org.altbeacon.beacon.logging.LogManager;
-import org.altbeacon.beacon.logging.Loggers;
-import org.altbeacon.beacon.service.BeaconService;
-import org.altbeacon.beacon.simulator.BeaconSimulator;
-import org.altbeacon.beacon.service.StartRMData;
-
 import android.annotation.TargetApi;
 import android.bluetooth.BluetoothManager;
 import android.content.ComponentName;
@@ -49,6 +34,23 @@ import android.os.IBinder;
 import android.os.Message;
 import android.os.Messenger;
 import android.os.RemoteException;
+
+import org.altbeacon.beacon.logging.LogManager;
+import org.altbeacon.beacon.logging.Loggers;
+import org.altbeacon.beacon.service.BeaconService;
+import org.altbeacon.beacon.service.RangeState;
+import org.altbeacon.beacon.service.RangedBeacon;
+import org.altbeacon.beacon.service.RunningAverageRssiFilter;
+import org.altbeacon.beacon.service.StartRMData;
+import org.altbeacon.beacon.simulator.BeaconSimulator;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * An class used to set up interaction with beacons from an <code>Activity</code> or <code>Service</code>.
@@ -664,6 +666,36 @@ public class BeaconManager {
 
     public static void setDistanceModelUpdateUrl(String url) {
         distanceModelUpdateUrl = url;
+    }
+
+    /**
+     * Default class for rssi filter/calculation implementation
+     */
+    protected static Class rssiFilterImplClass = RunningAverageRssiFilter.class;
+
+    public static void setRssiFilterImplClass(Class c) {
+        rssiFilterImplClass = c;
+    }
+
+    public static Class getRssiFilterImplClass() {
+        return rssiFilterImplClass;
+    }
+
+    /**
+     * Allow the library to use a tracking cache
+     * @param useTrackingCache
+     */
+    public static void setUseTrackingCache(boolean useTrackingCache) {
+        RangeState.setUseTrackingCache(useTrackingCache);
+    }
+
+    /**
+     * Set the period of time, in which a beacon did not receive new
+     * measurements
+     * @param maxTrackingAge in milliseconds
+     */
+    public void setMaxTrackingAge(int maxTrackingAge) {
+        RangedBeacon.setMaxTrackinAge(maxTrackingAge);
     }
 
     public static void setBeaconSimulator(BeaconSimulator beaconSimulator) {
--- a/src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
+++ b/src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
@@ -7,10 +7,13 @@ import org.altbeacon.beacon.logging.LogManager;
  * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
  * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
  * Note: a smoother average needs longer to "settle down"
+ * Note: For signals, that change rather frequently (say, 1Hz or faster) and tend to vary more a recommended
+ *       value would be 0,1 (that means the actual value is changed by 10% of the difference between the
+ *       actual measurement and the actual average)
+ *       For signals at lower rates (10Hz) a value of 0.25 to 0.5 would be appropriate
  */
 public class ArmaRssiFilter implements RssiFilter {
 
-    private static int DEFAULT_ARMA_MEASUREMENT = -75;  //What is the typical RSSI that beacons are sending with?
     private static double DEFAULT_ARMA_SPEED = 0.1;     //How likely is it that the RSSI value changes?
                                                         //Note: the more unlikely, the higher can that value be
                                                         //      also, the lower the (expected) sending frequency,
@@ -18,16 +21,21 @@ public class ArmaRssiFilter implements RssiFilter {
 
     private static final String TAG = "ArmaRssiFilter";
     //initially set to min value
-    private int armaMeasurement = -75;
+    private int armaMeasurement;
     private double armaSpeed = 0.1;
+    private boolean isInitialized = false;
 
     public ArmaRssiFilter() {
-        this.armaMeasurement = DEFAULT_ARMA_MEASUREMENT;
         this.armaSpeed = DEFAULT_ARMA_SPEED;
     }
 
 	public void addMeasurement(Integer rssi) {
         LogManager.d(TAG, "adding rssi: %s", rssi);
+        //use first measurement as initialization
+        if (!isInitialized) {
+            armaMeasurement = rssi;
+            isInitialized = true;
+        };
         armaMeasurement = Double.valueOf(armaMeasurement - armaSpeed * (armaMeasurement - rssi)).intValue();
         LogManager.d(TAG, "armaMeasurement: %s", armaMeasurement);
 	}
@@ -40,10 +48,6 @@ public class ArmaRssiFilter implements RssiFilter {
 		return armaMeasurement;
 
 	}
-
-    public static void setDEFAULT_ARMA_MEASUREMENT(int default_arma_measurement) {
-        DEFAULT_ARMA_MEASUREMENT = default_arma_measurement;
-    }
 
     public static void setDEFAULT_ARMA_SPEED(double default_arma_speed) {
         DEFAULT_ARMA_SPEED = default_arma_speed;
--- a/src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
+++ b/src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
@@ -2,7 +2,7 @@ package org.altbeacon.beacon.service;
 
 import org.altbeacon.beacon.logging.LogManager;
 
-/*
+/**
  * This filter calculates its rssi on base of an auto regressive moving average (ARMA)
  * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
  * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
@@ -10,13 +10,25 @@ import org.altbeacon.beacon.logging.LogManager;
  */
 public class ArmaRssiFilter implements RssiFilter {
 
+    private static int DEFAULT_ARMA_MEASUREMENT = -75;  //What is the typical RSSI that beacons are sending with?
+    private static double DEFAULT_ARMA_SPEED = 0.1;     //How likely is it that the RSSI value changes?
+                                                        //Note: the more unlikely, the higher can that value be
+                                                        //      also, the lower the (expected) sending frequency,
+                                                        //      the higher should that value be
+
     private static final String TAG = "ArmaRssiFilter";
     //initially set to min value
-    private int armaMeasurement = -100;
+    private int armaMeasurement = -75;
+    private double armaSpeed = 0.1;
+
+    public ArmaRssiFilter() {
+        this.armaMeasurement = DEFAULT_ARMA_MEASUREMENT;
+        this.armaSpeed = DEFAULT_ARMA_SPEED;
+    }
 
 	public void addMeasurement(Integer rssi) {
         LogManager.d(TAG, "adding rssi: %s", rssi);
-        armaMeasurement = Double.valueOf(armaMeasurement - 0.1*(armaMeasurement - rssi)).intValue();
+        armaMeasurement = Double.valueOf(armaMeasurement - armaSpeed * (armaMeasurement - rssi)).intValue();
         LogManager.d(TAG, "armaMeasurement: %s", armaMeasurement);
 	}
 
@@ -24,10 +36,17 @@ public class ArmaRssiFilter implements RssiFilter {
         return false;
 	}
 
-	public double calculateRunningAverage() {
+	public double calculateRssi() {
 		return armaMeasurement;
 
 	}
-	
+
+    public static void setDEFAULT_ARMA_MEASUREMENT(int default_arma_measurement) {
+        DEFAULT_ARMA_MEASUREMENT = default_arma_measurement;
+    }
+
+    public static void setDEFAULT_ARMA_SPEED(double default_arma_speed) {
+        DEFAULT_ARMA_SPEED = default_arma_speed;
+    }
 	
 }
--- a/src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
+++ b/src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
@@ -1,0 +1,33 @@
+package org.altbeacon.beacon.service;
+
+import org.altbeacon.beacon.logging.LogManager;
+
+/*
+ * This filter calculates its rssi on base of an auto regressive moving average (ARMA)
+ * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
+ * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
+ * Note: a smoother average needs longer to "settle down"
+ */
+public class ArmaRssiFilter implements RssiFilter {
+
+    private static final String TAG = "ArmaRssiFilter";
+    //initially set to min value
+    private int armaMeasurement = -100;
+
+	public void addMeasurement(Integer rssi) {
+        LogManager.d(TAG, "adding rssi: %s", rssi);
+        armaMeasurement = Double.valueOf(armaMeasurement - 0.1*(armaMeasurement - rssi)).intValue();
+        LogManager.d(TAG, "armaMeasurement: %s", armaMeasurement);
+	}
+
+	public boolean noMeasurementsAvailable() {
+        return false;
+	}
+
+	public double calculateRunningAverage() {
+		return armaMeasurement;
+
+	}
+	
+	
+}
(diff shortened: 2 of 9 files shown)


davidgyoung:
@IgorZaus, I tested this and it performs as expected, nice work!  I do have a number of comments about some particulars.  Please take a look and let me know what you think.  

In general, I think the default settings for the ArmaRssiFilter may not work well in some situations.   If somebody turns on a beacon transmitting at 1Hz with a measured RSSI of -50, it will take about 20 seconds for the algorithm to return a value that is within 10% of the actual RSSI.  See my calculations below.  I have specific comments inline that may help address this.

```
t=0s : -100-0.1(-100-(-50)) = -95
t=1s : -95-0.1(-95-(-50)) = -90.5
t=2s : -90.5-0.1(-90.5-(-50) = -86.45
t=3s : -82.805
t=4s : -79.5245
t=5s : -73.914845
...
t=20s : -54.923854510918055
```

src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -666,6 +668,36 @@ public static void setDistanceModelUpdateUrl(String url) {
         distanceModelUpdateUrl = url;
     }
 
+    /**
+     * Default class for rssi filter/calculation implementation
+     */
+    protected static Class rssiFilterImplClass = RunningAverageRssiFilter.class;
```
davidgyoung: instead of Class, can we make this RssiFilter?  (Same goes for the getter and setter)

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
```diff
@@ -0,0 +1,33 @@
+package org.altbeacon.beacon.service;
+
+import org.altbeacon.beacon.logging.LogManager;
+
+/*
```
davidgyoung: should be a javadoc class comment (start with /**)

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
```diff
@@ -0,0 +1,33 @@
...
+ * This filter calculates its rssi on base of an auto regressive moving average (ARMA)
+ * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
+ * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
+ * Note: a smoother average needs longer to "settle down"
+ */
+public class ArmaRssiFilter implements RssiFilter {
```
davidgyoung: Would be good to create a `ArmaRssiFilterTest` class to do a unit test of this.  You can see some examples in the test folder.  You run them with `./gradlew test`

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private static final String TAG = "ArmaRssiFilter";
+    //initially set to min value
+    private int armaMeasurement = -100;
```
davidgyoung: Should the initial value be -100?  That is a particularly weak signal.  See my general comment about how long it takes for a -50 rssi signal to settle out without 10% of the expected value on a 1Hz transmitter.  Should this be configurable?

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
```diff
@@ -0,0 +1,33 @@
...
+
+	public void addMeasurement(Integer rssi) {
+        LogManager.d(TAG, "adding rssi: %s", rssi);
+        armaMeasurement = Double.valueOf(armaMeasurement - 0.1*(armaMeasurement - rssi)).intValue();
```
davidgyoung: should the 0.1 be a configurable value?  perhaps a static class member.  Should the default value be bigger (per general issue)

src/main/java/org/altbeacon/beacon/service/RangeState.java
```diff
@@ -54,7 +52,8 @@ public void addBeacon(Beacon beacon) {
             rangedBeacon.updateBeacon(beacon);
         }
         else {
-            LogManager.d(TAG, "adding %s to new rangedBeacon", beacon);
+            //TODO: make the used implementation an option
```
davidgyoung: This is done, right?

src/main/java/org/altbeacon/beacon/service/RangeState.java
```diff
@@ -54,7 +52,8 @@ public void addBeacon(Beacon beacon) {
...
+            LogManager.d(TAG, "adding %s to new rangedBeacon (ARMA)", beacon);
```
davidgyoung: This is not always using AMRA, so this should not be in the comment, right?

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
```diff
@@ -1,23 +1,30 @@
...
+import java.lang.reflect.Constructor;
+
 public class RangedBeacon {
-	private static final String TAG = "RangedBeacon";
-	public static long DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS = 20000; /* 20 seconds */
-	private static long sampleExpirationMilliseconds = DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS;
+
+    private static final String TAG = "RangedBeacon";
+    public static long DEFAULT_MAX_TRACKING_AGE = 5000; /* 5 Seconds */
+    public static long maxTrackinAge = DEFAULT_MAX_TRACKING_AGE; /* 5 Seconds */
```
davidgyoung: maxTrackinAge -> maxTrackingAge (same change for accessor methods)

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
```diff
@@ -1,23 +1,30 @@
...
+    protected RssiFilter filter = null;
 
 	public RangedBeacon(Beacon beacon) {
-		updateBeacon(beacon);
+        //set RSSI filter
+        try {
+            Constructor cons = BeaconManager.getRssiFilterImplClass().getConstructors()[0];
+            filter = (RssiFilter)cons.newInstance();
+        } catch (Exception e) {
+            e.printStackTrace();
```
davidgyoung: Should do a LogManager.e here

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
```diff
@@ -1,23 +1,30 @@
...
-	private static final String TAG = "RangedBeacon";
-	public static long DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS = 20000; /* 20 seconds */
-	private static long sampleExpirationMilliseconds = DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS;
+
+    private static final String TAG = "RangedBeacon";
+    public static long DEFAULT_MAX_TRACKING_AGE = 5000; /* 5 Seconds */
+    public static long maxTrackinAge = DEFAULT_MAX_TRACKING_AGE; /* 5 Seconds */
     private boolean mTracked = true;
-    private ArrayList<Measurement> mMeasurements = new ArrayList<Measurement>();
+    protected long lastTracked = 0;
```
davidgyoung: Perhaps rename to lastTrackedTimeMillis

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
```diff
@@ -49,65 +56,26 @@ public void commitMeasurements() {
         }
     }
 
-	public static void setSampleExpirationMilliseconds(long milliseconds) {
```
davidgyoung: setSampleExpirationMilliseconds needs to be retained for backwards compatibility.  I suggest we change it to check to see if the filter is an instance of RunningAverageRssiFilter, and if so, call a method to set an equivalent value on the class.  If it is not, log a warning saying this will do nothing.

src/main/java/org/altbeacon/beacon/service/RssiFilter.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Interface that can be implemented to overwrite measurement and filtering
+ * of RSSI values
+ */
+public interface RssiFilter {
+
+    public void addMeasurement(Integer rssi);
+    public boolean noMeasurementsAvailable();
+    public double calculateRunningAverage();
```
davidgyoung: Perhaps make more generic like: calculateFilteredRssi()

davidgyoung: Perhaps make more generic like: calculateFilteredRssi()

src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
```diff
@@ -0,0 +1,78 @@
+package org.altbeacon.beacon.service;
+
+import org.altbeacon.beacon.logging.LogManager;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.Iterator;
+
+/**
```
davidgyoung: Javadoc should say more about what this does like the javadoc for ArmaRssiFilter

src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
```diff
@@ -0,0 +1,78 @@
...
+ * Created by AS on 17.02.2015.
+ */
+public class RunningAverageRssiFilter implements RssiFilter {
```
davidgyoung: I know this is not your code, but now that it is a separate class, it would be great to have a `RunningAverageRssiFilterTest` to add test coverage for this, too

IgorZaus:
Yes, you’re right – I’ll make the default value „-75” and make both values configurable!

 

Von: David G. Young [mailto:notifications@github.com] 
Gesendet: Mittwoch, 25. Februar 2015 01:15
An: AltBeacon/android-beacon-library
Cc: IgorZaus
Betreff: Re: [android-beacon-library] Changes for #113 and #114 (#128)

 

@IgorZaus <https://github.com/IgorZaus> , I tested this and it performs as expected, nice work! I do have a number of comments about some particulars. Please take a look and let me know what you think. 

In general, I think the default settings for the ArmaRssiFilter may not work well in some situations. If somebody turns on a beacon transmitting at 1Hz with a measured RSSI of -50, it will take about 20 seconds for the algorithm to return a value that is within 10% of the actual RSSI. See my calculations below. I have specific comments inline that may help address this.

t=0s : -100-0.1(-100-(-50)) = -95
t=1s : -95-0.1(-95-(-50)) = -90.5
t=2s : -90.5-0.1(-90.5-(-50) = -86.45
t=3s : -82.805
t=4s : -79.5245
t=5s : -73.914845
...
t=20s : -54.923854510918055

src/main/java/org/altbeacon/beacon/BeaconManager.java
IgorZaus: I'm not sure about this - we need to create an instance in runtime, because every beacon needs its own filter object. Can you clarify, how you would change this?

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
IgorZaus: Yes, of course, I changed this!

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
IgorZaus: Yes, created a first test. More tests will follow.

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
IgorZaus: I made this "-75" by default and both values configurable (with static class methods)

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
IgorZaus: see comment above. I wouldn't make this value much bigger, because otherwise the average is too "unstable". Although, I guess the RSSI value of a beacon is unlikely to change, right? In that case one could set this to "0.5".

src/main/java/org/altbeacon/beacon/service/RangeState.java
IgorZaus: Yes, removed it

src/main/java/org/altbeacon/beacon/service/RangeState.java
IgorZaus: Yes, changed it back to original comment (I used this during testing)

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
IgorZaus: Changed

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
IgorZaus: Done.

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
IgorZaus: Done

src/main/java/org/altbeacon/beacon/service/RangedBeacon.java
IgorZaus: I see... changed this, too. But need to store the value of "sampleExpirationMilliseconds" in RangedBeacon (because its set in a static method) and transfer it to the filter on instantiation

src/main/java/org/altbeacon/beacon/service/RssiFilter.java
IgorZaus: Changed it to "calculateRssi" - which is what it actually does :-)

src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
IgorZaus: Ok, did this - you might want to check the explanation, though -:)

src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
IgorZaus: Ok, let's see if I can think of some useful tests on this

src/main/java/org/altbeacon/beacon/BeaconManager.java
davidgyoung: Yes, I I just mean use RssiFilter instead of Class like this:

        protected static RssiFilter rssiFilterImplClass = RunningAverageRssiFilter.class;

davidgyoung:
One other idea... What if the ArmaRssiFilter does not have an initial value, and instead initializes itself with the first RSSI value reported to it?  This would make it converge to a stable value faster regardless of how the measured RSSI differs from its initial value.

src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * This filter calculates its rssi on base of an auto regressive moving average (ARMA)
+ * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
+ * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
+ * Note: a smoother average needs longer to "settle down"
+ */
+public class ArmaRssiFilter implements RssiFilter {
+
+    private static int DEFAULT_ARMA_MEASUREMENT = -75;  //What is the typical RSSI that beacons are sending with?
+    private static double DEFAULT_ARMA_SPEED = 0.1;     //How likely is it that the RSSI value changes?
```
davidgyoung: This constant is probably a good value for applications where the mobile device is generally stationary relative to the beacon.  But in some applications, like indoor navigation, the Mobile device is typically moving at velocities of 1m/s.  For such applications, it is important to have the RSSI respond to updates more quickly, so a value of 1.0 may be more appropriate.  I don't have strong opinions about which should be the default, but the JavaDocs should probably explain this and suggest an alternate value for use cases where changes in RSSI need to come more quickly.

Another factor in how quickly the RSSI will change using this algorithm is the transmission frequency of the beacon.  Non-connectable beacons transmitting at 10Hz will allow RSSI changes 10x faster than ones transmitting at 1Hz.

IgorZaus: No, a value of 1 would make the average directly follow the measurement – thus, we wouldn’t have any averaging effect. I think, if we change the initial value to the first measurement we should get a good result. I will add some notes here to JavaDoc – also addressing cases, where a value of 0.5 might be appropriate. 

Best, André

 

Von: David G. Young [mailto:notifications@github.com] 
Gesendet: Mittwoch, 4. März 2015 14:31
An: AltBeacon/android-beacon-library
Cc: IgorZaus
Betreff: Re: [android-beacon-library] Changes for #113 and #114 (#128)

 

In src/main/java/org/altbeacon/beacon/service/ArmaRssiFilter.java <https://github.com/AltBeacon/android-beacon-library/pull/128#discussion_r25773860> :

> @@ -0,0 +1,52 @@
> +package org.altbeacon.beacon.service;
> +
> +import org.altbeacon.beacon.logging.LogManager;
> +
> +/**
> + * This filter calculates its rssi on base of an auto regressive moving average (ARMA)
> + * It needs only the current value to do this; the general formula is  n(t) = n(t-1) - c * (n(t-1) - n(t))
> + * where c is a coefficient, that denotes the smoothness - the lower the value, the smoother the average
> + * Note: a smoother average needs longer to "settle down"
> + */
> +public class ArmaRssiFilter implements RssiFilter {
> +
> +    private static int DEFAULT_ARMA_MEASUREMENT = -75;  //What is the typical RSSI that beacons are sending with?
> +    private static double DEFAULT_ARMA_SPEED = 0.1;     //How likely is it that the RSSI value changes?

This constant is probably a good value for applications where the mobile device is generally stationary relative to the beacon. But in some applications, like indoor navigation, the Mobile device is typically moving at velocities of 1m/s. For such applications, it is important to have the RSSI respond to updates more quickly, so a value of 1.0 may be more appropriate. I don't have strong opinions about which should be the default, but the JavaDocs should probably explain this and suggest an alternate value for use cases where changes in RSSI need to come more quickly.

Another factor in how quickly the RSSI will change using this algorithm is the transmission frequency of the beacon. Non-connectable beacons transmitting at 10Hz will allow RSSI changes 10x faster than ones transmitting at 1Hz.

src/main/java/org/altbeacon/beacon/BeaconManager.java
IgorZaus: This gives me a type error – IntelliJ or Java (?) is not able to do or allow that casting (from specific class to a type)

I get: Incompatible types.

 

Can you please check!?

 

Von: David G. Young [mailto:notifications@github.com] 
Gesendet: Mittwoch, 4. März 2015 14:01
An: AltBeacon/android-beacon-library
Cc: IgorZaus
Betreff: Re: [android-beacon-library] Changes for #113 and #114 (#128)

 

In src/main/java/org/altbeacon/beacon/BeaconManager.java <https://github.com/AltBeacon/android-beacon-library/pull/128#discussion_r25772294> :

> @@ -666,6 +668,36 @@ public static void setDistanceModelUpdateUrl(String url) {
>          distanceModelUpdateUrl = url;
>      }
>  
> +    /**
> +     * Default class for rssi filter/calculation implementation
> +     */
> +    protected static Class rssiFilterImplClass = RunningAverageRssiFilter.class;

Yes, I I just mean use RssiFilter instead of Class like this:

    protected static RssiFilter rssiFilterImplClass = RunningAverageRssiFilter.class;

IgorZaus:
Very good idea! I implemented this. Actually, in signal processing, it’s done exactly that way – don’t know, how I forgot/missed this…

 

Von: David G. Young [mailto:notifications@github.com] 
Gesendet: Mittwoch, 4. März 2015 14:10
An: AltBeacon/android-beacon-library
Cc: IgorZaus
Betreff: Re: [android-beacon-library] Changes for #113 and #114 (#128)

 

One other idea... What if the ArmaRssiFilter does not have an initial value, and instead initializes itself with the first RSSI value reported to it? This would make it converge to a stable value faster regardless of how the measured RSSI differs from its initial value.

src/main/java/org/altbeacon/beacon/BeaconManager.java
davidgyoung: Sorry, forget this comment.  This suggestion won't work because the code sets the Class instance, not an Object instance, and only Object instances can use the Interface as a type.  Let's just leave it as-is.

davidgyoung:
I need to do some more testing on this then I think it is ready to merge.

samskiter:
Hi, I was just playing with `RunningAverageRssiFilter.setSampleExpirationMilliseconds()` and I'm not convinced it works. Won't the `RangedBacon` constructor write over the top of it every time?

IgorZaus:
Hi Sam, I will check this ASAP… I made this for a prototype that was shown in April. It seemed to worked back than, but it never made it into a product, so there was no further testing…

Kind regards, André

 

Von: Sam Duke [mailto:notifications@github.com] 
Gesendet: Dienstag, 29. September 2015 17:13
An: AltBeacon/android-beacon-library <android-beacon-library@noreply.github.com>
Cc: IgorZaus <zaus@sektor7g.de>
Betreff: Re: [android-beacon-library] Changes for #113 and #114 (#128)

 

Hi, I was just playing with RunningAverageRssiFilter.setSampleExpirationMilliseconds() and I'm not convinced it works. Won't the RangedBacon constructor write over the top of it every time?

samskiter:
Looks like it's to do with the 'backward compatibility' that was introduced.

Current solution isn't so bad.

```
RunningAverageRssiFilter.setSampleExpirationMilliseconds(5000L);
RangedBeacon.setSampleExpirationMilliseconds(5000L);
```

Issue #134, opened 2015-02-27
Motorola devices allow transmission if isMultipleAdvertisementSupported returns false


davidgyoung:
Currently, the `BeaconTransmitter.checkTransmissionSupported()` method returns an error code if BluetoothAdapter.isMultipleAdvertisementSupported() returns false.  This is proper behavior for Nexus 9 and Nexus 6 devices with Android 5.0, because the getBluetoothLeAdvertiser() function in the Android source code (BluetoothAdapter.java) contains :

    if (!isMultipleAdvertisementSupported()) {
                return null;
        }

However, in the builds for Motorola devices as of 5.0.2, this reportedly has been changed to return an advertiser even if isMultipleAdvertisementSupported() returns false.

The `BeaconTransmitter.checkTransmissionSupported()` method therefore needs to be changed to try to get an advertiser regardless of whether BluetoothAdapter.isMultipleAdvertisementSupported() returns false.

davidgyoung:
Fixed in 	f3d4337, and put into release 2.2 of Beacon Locate for Android.

PR #135, opened 2015-02-27
Allow beacon transmission on Motorola devices

davidgyoung:
This addresses issue #134, which appears to be blocking beacon transmission on Motorola devices.


mattwritescode:
The best changes seem to be short and simple like this one.

davidgyoung:
Our friend at Motorola has confirmed that this makes transmission work on Moto G, Moto X,  and the new Moto E with 5.x.  

I'm going to roll a new release.  Can anybody check the Moto G in the office and see if a 5.x upgrade is available?  It is a first gen Moto G, not a second gen,  so I am not 100% sure it will work.

davidhelms:
Sweet!

jnebeker:
Just checked the Moto G, no update available

davidgyoung:
Nuts

davidgyoung:
Based on initial testing, this does allow Motorola devices to advertise.

Issue #148, opened 2015-03-31
Callbacks are not made on the main thread, allowing state to be


samskiter:

davidgyoung:
I'm not sure I understand why not making callbacks on the main thread causes this issue.  This is by design, with the intention that the user needs to explicitly execute code on the main thread inside the callback.

I appreciate the attention, @samskiter.

samskiter:
It's a combination of calling on a background thread and having an activity as the receiver (albeit, indirectly via the anonymous inner class).

Consider the sequence of events:

1. beacon is found -> background callback on Activity A
2. Activity A calls dispatches to the main thread to make some UI change
3. Activity A undergoes a configuration change. Activity A is replaced on screen by Activity B. Activity A is still retained by the call made above in step 2
4. callback arrives at Activity A, which updates its UI. This is never seen by the user

If the UI change mentioned is to pop up a dialog or change any fragment, we will certainly see a crash unless commitAllowingStateLoss() was used (and the state will be lost, as per step 4)

One nice fact about config changes is that internally they happen as a single sequential block. Meaning that if you submit your callback to the UI thread you will definitely be one side of a config change or the other. This means you must also make sure you read your notifier on the main thread as the config change will likely involve changing the notifier.

Hope that clarifies?

davidgyoung:
One concern I have with this change is that if an app developer has a lot of code inside their callback (which they should not do anyway, but folks still do it), having it execute on the main thread might make the UI lag.   Pushing this change to app developers might require them to re-write their currently working code.

I understand that the current implementation allows the user to shoot themselves in the foot if they are getting callbacks inside an Activity and that Activity disappears.  Perhaps this should simply be solved by updating the reference app to demonstrate how to do this cleanly with the current implementation.

samskiter:
Hmm, the regression argument is important, you're right. The way to fix things up on the reference implementation would be use a retained fragment to hold the notifier and post to the main thread from there (where you then read the fragments "getActivity()" and pass the event through).

You could support different kinds of listeners/notifiers for the implementation and deprecate the old setNotifier endpoint... I think in general it would be great to have multiple notifiers/listeners, so implementing that at the same time might result in a change of method name, alerting existing devs to the change...

PR #149, opened 2015-03-31
Always deliver scan/ranging results on the main thread

samskiter:
UNTESTED, but hopefully not too complex.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconIntentProcessor.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconIntentProcessor.java
@@ -44,6 +44,19 @@ public class BeaconIntentProcessor extends IntentService {
 
 	@Override
 	protected void onHandleIntent(Intent intent) {
+        //Make sure we always deliver results on the main thread
+        if (Looper.getMainLooper().getThread() != Thread.currentThread()) {
+            Handler mainHandler = new Handler(Looper.getMainLooper());
+        
+            Runnable myRunnable = new Runnable() {
+                @Override
+                public void run() {
+                    onHandleIntent(intent)
+                }
+            };
+            mainHandler.post(myRunnable);
+            return;
+        }
 		LogManager.d(TAG, "got an intent to process");
 		
 		MonitoringData monitoringData = null;


davidgyoung:
As discussed in #148 this is a breaking change, so it cannot be merged until a 3.0 release.

samskiter:
LGTM... just to add another vote to this, I just hit:

`android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.`

After adjusting some ui from the `didRangeBeaconsInRegion` callback

marcosalis:
I agree that this is a valuable change to the library behaviour. I think that the most common user case for the callbacks is to perform operations that will eventually modify the UI, so it makes perfect sense to receive the callbacks from the main thread (and this is the approach that most well known third-party libraries take). If the caller needs to perform long running operations, it makes sense that they spawn their own Executors, or thread, or whatever the threading model of their application requires.
What's more, AltBeacon (as far as I could see in the documentation) doesn't specify if it's thread safe, so the first instinct would be to only access it from the main thread (and expect callbacks on the same thread).

src/main/java/org/altbeacon/beacon/BeaconIntentProcessor.java
```diff
@@ -44,6 +47,19 @@ public BeaconIntentProcessor() {
 
 	@Override
 	protected void onHandleIntent(Intent intent) {
+        //Make sure we always deliver results on the main thread
```
marcosalis: `onHandleIntent()` always gets executed on a worker thread, as this is the expected behaviour of an `IntentService` (see [documentation](https://developer.android.com/reference/android/app/IntentService.html#onHandleIntent(android.content.Intent))).

It's probably better not to mess with Android callbacks by calling them directly.
It might be safer to instantiate the Handler in the onCreate() and add a new method that posts to the main thread the calls to the listeners.

hashbrown:
BTW @davidgyoung, you've already unintentionally made this breaking change in 2.12 (for Android 8), where you use the BeaconLocalBroadcastProcesser to handle Range and Monitor notifications that are intents captured from a BroadcastReceiver.  Since the intents are always handled on the main thread, and BeaconLocalBoradcasProcessor is not an IntentService as before, all Ranging callbacks are actually received by the application on its main thread. 

Personally I'd like to see this fixed in the 2.X series since it is a breaking api change, and has at least been detrimental to my app.  Also, if this PR is accepted, can we please configure this behavior?  I do not want to receive these on the main thread.

davidgyoung:
I had not intended such a change, no.  At this point I agree that documentation of the existing behavior and configurability is probably the best option.

PR #153, opened 2015-04-04
Various improvements to Identifier, including UUID awareness

### LINKED ISSUE #150: Identifier: Convenient use of java.util.UUID
I just started using this library today and had a look at the implementation of `org.altbeacon.beacon.Identifier`. I'm curious why there's no compatibility with `java.util.UUID`. I see that `Identifier` supports identfication in form of arbitrarily long `byte[]` arrays, but still you have `String toUuidString`, so `UUID toUuid()` and `Identifier fromUuid(UUID)` do not seem too far fetched. What gives?

flowlo:
For detailed reasoning about those changes, please see comments directly in the change view.

Feedback is requested. I'd supply more tests and docs once this receives general acceptance.

Fixes #150

### DIFF
--- a/src/main/java/org/altbeacon/beacon/Identifier.java
+++ b/src/main/java/org/altbeacon/beacon/Identifier.java
@@ -1,7 +1,9 @@
 package org.altbeacon.beacon;
 
-import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.nio.LongBuffer;
 import java.util.Arrays;
+import java.util.UUID;
 import java.util.regex.Pattern;
 
 /**
@@ -14,16 +16,15 @@ import java.util.regex.Pattern;
  */
 public class Identifier implements Comparable<Identifier> {
     private static final Pattern HEX_PATTERN = Pattern.compile("^0x[0-9A-Fa-f]*$");
-    private static final Pattern DECIMAL_PATTERN = Pattern.compile("^[0-9]+$");
-    private static final Pattern UUID_PATTERN = Pattern.compile("^[0-9A-Fa-f]{8}-?[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{12}$");
-    private static final char[] hexArray = "0123456789abcdef".toCharArray();
-    private static final BigInteger maxInteger = BigInteger.valueOf(65535);
+    private static final Pattern DECIMAL_PATTERN = Pattern.compile("^[0-9]{1,5}$");
+    private static final Pattern UUID_PATTERN = Pattern.compile("^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$");
+    private static final int maxInteger = 65535;
 
     private final byte[] mValue;
 
     /**
      * Allowed formats:
-     * <ul><li>UUID: 2F234454-CF6D-4A0F-ADF2-F4911BA9FFA6 (16 bytes Identifier, dashes optional)
+     * <ul><li>UUID: 2F234454-CF6D-4A0F-ADF2-F4911BA9FFA6 (16 bytes Identifier)
      *   <li>HEX: 0x000000000003 (number of bytes is based on String length)
      *   <li>Decimal: 65536 (2 bytes Identifier)</ul>
      * @param stringValue string to parse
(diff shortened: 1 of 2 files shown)


src/main/java/org/altbeacon/beacon/Identifier.java
```diff
@@ -14,16 +16,15 @@
...
-    private static final BigInteger maxInteger = BigInteger.valueOf(65535);
+    private static final Pattern DECIMAL_PATTERN = Pattern.compile("^[0-9]{1,5}$");
+    private static final Pattern UUID_PATTERN = Pattern.compile("^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$");
+    private static final int maxInteger = 65535;
 
     private final byte[] mValue;
 
     /**
      * Allowed formats:
-     * <ul><li>UUID: 2F234454-CF6D-4A0F-ADF2-F4911BA9FFA6 (16 bytes Identifier, dashes optional)
```
flowlo: Dashes are *not* optional, according to [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt) (ABNF in section 3 on page 3). This also applies to the changes to `UUID_PATTERN`.

src/main/java/org/altbeacon/beacon/Identifier.java
```diff
@@ -14,16 +16,15 @@
  */
 public class Identifier implements Comparable<Identifier> {
     private static final Pattern HEX_PATTERN = Pattern.compile("^0x[0-9A-Fa-f]*$");
-    private static final Pattern DECIMAL_PATTERN = Pattern.compile("^[0-9]+$");
-    private static final Pattern UUID_PATTERN = Pattern.compile("^[0-9A-Fa-f]{8}-?[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{4}-?[0-9A-Fa-f]{12}$");
-    private static final char[] hexArray = "0123456789abcdef".toCharArray();
-    private static final BigInteger maxInteger = BigInteger.valueOf(65535);
```
flowlo: `65535 < Integer.MAX_VALUE` can be assumed nearly everywhere. Therefore `BigInteger` is not needed and we can use an ordinary `int`.

src/main/java/org/altbeacon/beacon/Identifier.java
```diff
@@ -14,16 +16,15 @@
...
+    private static final Pattern DECIMAL_PATTERN = Pattern.compile("^[0-9]{1,5}$");
```
flowlo: Make use of the interval operator to enforce a limit already in the regular expression.

src/main/java/org/altbeacon/beacon/Identifier.java
```diff
@@ -149,7 +170,7 @@ public String toString() {
             return Integer.toString(toInt());
         }
         if (mValue.length == 16) {
-            return toUuidString();
```
flowlo: As `toUuidString()` is now deprecated, stop using it here ...

src/main/java/org/altbeacon/beacon/Identifier.java
```diff
@@ -225,13 +246,11 @@ public boolean equals(Object that) {
     }
 
     private static String toHexString(byte[] bytes, int start, int length) {
-        char[] hexChars = new char[length * 2];
-        for ( int i = 0; i < length; i++ ) {
-            int v = bytes[start + i] & 0xFF;
-            hexChars[i * 2] = hexArray[v >>> 4];
-            hexChars[i * 2 + 1] = hexArray[v & 0x0F];
+        StringBuilder sb = new StringBuilder();
```
flowlo: `StringBuilder`s are made to build strings and have format strings.

src/main/java/org/altbeacon/beacon/Identifier.java
davidgyoung: In order to keep this from being a breaking change to the API, parsing uuids without dashes must be allowed.  I am OK with removing this from the docs, however to discourage it.

src/main/java/org/altbeacon/beacon/Identifier.java
```diff
@@ -225,13 +246,11 @@ public boolean equals(Object that) {
...
+        for (int i = 0; i < length; i++) {
+            sb.append(String.format("%02x", bytes[start + i]));
```
davidgyoung: Great change.  Don't know why I did not do this.

davidgyoung:
Looks good, @flowlo.  I would add a couple of tests:

1. make sure uuids parse without dashes for backward compatibility with existing apps that are already doing this.
2. Test that integer identifiers over 65535 are not allowed.

flowlo:
Thank you for the feedback! Expect an update later today.

flowlo:
Should be good to merge now. :sparkles:

Summary:

 * Add `toUuid()` and `fromUuid()` (very cheap, no strings involved, uses type-safe `java.util.UUID`)
 * Deprecate `toUuidString()` in favour of `toUuid()`
 * Remove unneeded dependency on `BigInteger` and make the regular expression for decimal numbers a bit more clever to rule out huge numbers before even sending them to `Integer.valueOf`
 * `s/maxInteger/MAX_INTEGER` cause it's `static final`
 * Use `StringBuilder` instead of custom logic in `toHexString()`
 * Fix a bug in `toHexString()` that would make the result not be accepted by `parse()` because of missing `0x` prefix (as stated in the JavaDoc).
 * Add `toByteArray()` to be able to get raw bytes out of an identifier directly.
 * Add tests

For the next major release, it would be great if you could drop invalid UUIDs being parsed!

davidgyoung:
Looks good to me.

PR #166, opened 2015-04-15
v3: Remove convenience accessors, fix UUID parsing

lorenzleutgeb:
This is my wishlist for the next major version (breaking changes).

Convenience access for the first three Identifiers in a Region is complete madness. It provokes off-by-ones (consider `getId1()` vs. `getIdentifier(0)`) and implies that the first three Identifiers are somewhat special when they are not. My solution is removal of those. The constructor for three Identifiers now uses variadic arguments, so you can use it like you always did (just pass three Identifiers, no need to change any code!), but is more flexible.

UUIDs are now implemented correctly and do not maintain the backcompat bug (dashes).

Expect more to come, and please leave comments!

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
@@ -162,10 +162,9 @@ public class BeaconTransmitter {
             byteString += String.format("%02X", advertisingBytes[i]);
             byteString += " ";
         }
-        LogManager.d(TAG, "Starting advertising with ID1: %s ID2: %s ID3: %s and data: %s of size "
-                        + "%s", mBeacon.getId1(),
-                        mBeacon.getIdentifiers().size() > 1 ? mBeacon.getId2() : "",
-                        mBeacon.getIdentifiers().size() > 2 ? mBeacon.getId3() : "", byteString,
+        LogManager.d(TAG, "Starting advertising with %s and data: %s of size %d",
+                mBeacon,
+                byteString,
                 advertisingBytes.length);
 
         try{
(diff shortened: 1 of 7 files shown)


davidgyoung:
Thanks for the ideas, @flowlo.  

I can partly understand the hating on the convenience accessors for Id1, Id2 and Id3.  These are only special because they are the identifiers for the AltBeacon beacon type.  This beacon type actually has its own class `AltBeacon` where these identifier accessors could reside, but then that would require converting a Beacon to an AltBeacon first, which isn't very convenient.  I'm certainly open to ideas to make this better, but I do think it is good to offer a convenient way to access the three primary identifiers for use on the (currently) most common beacon type.

flowlo:
@davidgyoung I'd like to better understand the need for those convenience accessors. How are they related to AltBeacon beacons, what makes them special? I had a look at [the spec](https://github.com/AltBeacon/spec) but it does not talk about three identifiers. From a generic point of view, they don't make sense, but in case AltBeacon gives meaning to those very three IDs, it may be very valid to have them stick around.

Also, I have not touched the `Beacon` class, only `Identifier` (when it comes to accessors), so how does that relate anyway? Sorry, I'm a bit confused.

davidgyoung:
Right, this PR actually only removes the three convenience accessors on `Region` but they are the same ones as on `Beacon`.  

The reason Id1, Id2 and Id3 are special is because AltBeacons are typically used with identifiers that are interoperable with the iBeacon format, meaning the first identifier is a UUID, the second and third identifiers are 32-bit integers (called major and minor on iBeacon).  I'd say 90% plus of users of this library use Beacons with identifiers in these formats, so it is a common shorthand to use region.getId1() to get the UUID, region.getId2() to get the major and region.getId3() to get the minor.

Issue #155, opened 2015-04-05
Code goes spaghetti if viewed with tab width != 4


flowlo:
I just attempted to read `Beacon.java` and `Region.java` but totally went nuts as soon as I encountered that mess of whitespace.

*Tabs vs. Spaces* is an endless discussion. TBH I don't want to bring it up. This issue is about consistency and valid for both models: I see that most code uses 4 spaces as one level of indentation, but every now and then tabs kick in, which is suboptimal.

I personally choose tab width dependent on the language I read, and using a tab width of 2 here (really anything else than 4) will freak the reader out.

![Screenshot of Beacon.java with tab with of 2](https://cloud.githubusercontent.com/assets/542154/6997869/be0f61dc-dbcd-11e4-8305-704ee4de6a47.png)

This should be normalised throughout the whole codebase and have tests to cover future work. I'm happy to do this in case you want it to be done.

davidgyoung:
Yes, we should be standardized on a tab length of four and use spaces instead of the tab character.  This is following the Android code style guidelines.

samskiter:
+1

flowlo:
@samskiter Feel free to have a look at #167

davidgyoung:
Fixed in #167

PR #167, opened 2015-04-15
Format code

flowlo:
Should solve #155 

Majority of the changes here were done by doing `s/\t/    /` and `s/ +\n/\n/`, so nothing special. I also added and removed a few empty lines here and there I think.


samskiter:
LGTM

PR #171, opened 2015-04-17
Add Multiple Notifiers Support (Range and Monitor). Add Region Notifiers...

Maxeee09:
Yesterday, we spoke about adding multiple notifiers support. So I added this capability into my fork. Feel free to merge these changes if you want. Thank you for your work around beacons !

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -108,9 +108,9 @@ public class BeaconManager {
     protected static BeaconManager client = null;
     private final ConcurrentMap<BeaconConsumer, ConsumerInfo> consumers = new ConcurrentHashMap<BeaconConsumer,ConsumerInfo>();
     private Messenger serviceMessenger = null;
-    protected RangeNotifier rangeNotifier = null;
     protected RangeNotifier dataRequestNotifier = null;
-    protected MonitorNotifier monitorNotifier = null;
+    private final ArrayList<RangeNotifier> rangeNotifiers = new ArrayList<RangeNotifier>();
+    private final ArrayList<MonitorNotifier> monitorNotifiers = new ArrayList<MonitorNotifier>();
     private final ArrayList<Region> monitoredRegions = new ArrayList<Region>();
     private final ArrayList<Region> rangedRegions = new ArrayList<Region>();
     private final ArrayList<BeaconParser> beaconParsers = new ArrayList<BeaconParser>();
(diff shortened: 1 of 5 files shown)


src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -387,35 +387,93 @@ public boolean isBackgroundModeUninitialized() {
...
+     * allow to set Range Notifier
+     * Backward compatibility
      * @param notifier
-     * @see RangeNotifier
      */
-    public void setRangeNotifier(RangeNotifier notifier) {
-        rangeNotifier = notifier;
+    public void setRangeNotifier( RangeNotifier notifier )
+    {
+        addRangeNotifier( notifier );
```
davidgyoung: This could actually be backward compatible with the existing API if this method (and the monitoring equivalent) simply replaced the first notifier.

src/main/java/org/altbeacon/beacon/notifier/RegionMonitorNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.altbeacon.beacon.MonitorNotifier;
+import org.altbeacon.beacon.Region;
+
+import java.util.ArrayList;
+
+/**
+ * Created by maxence on 17/04/15.
+ */
+public abstract class RegionMonitorNotifier extends ArrayList<Region> implements MonitorNotifier
```
davidgyoung: Interesting implementation.   This would allow the RegionMonitorNotifier to exist alongside the existing MonitorNotifier.

davidgyoung: One consequence of this implementation is that the Regions list could be changed after monitoring is started.  And I believe the contains() method is not thread safe.  This could cause a ConcurrentModificationException if this method is called while code in another thread modifies the Regions in the list.

Maxeee09:
Yes indeed, I understand these two points. First I will enhance backward compatibility by removing old notifiers on new notifier set. And then I will change the RegionNotifier implementation to be thread safe. Thank you for your advices.

furkanvarol:
I believe this functionality should be added to library but this PR includes lots of changes due to change of code style. I humbly think this PR should be revisioned to keep library in same code style.

davidgyoung:
@furkanvarol, what do yuou mean by revisioned?

furkanvarol:
I mean this PR should be refactored to keep same code style in the library. For basic example, check
https://github.com/AltBeacon/android-beacon-library/pull/171/files?diff=unified#diff-d0a412658873761c6d450c9a9f8e0162L133.
Nevertheless, this PR does not actually change `setDebug` method but it changes its style like rest of the file. And this also causes code style difference within the library. Also, new files should follow the code style.

Btw, a style checker would be nice for the library or at least a guide like http://checkstyle.sourceforge.net/.

davidgyoung:
@Maxeee09, do you want to try updating the code style (whitespace, etc.) to match, per @furkanvarol's comment?

davidgyoung:
The easiest way might be to look at the diff  in this PR and use it as a guide to re-implementing the same code changes without modifying the style of the code, which tries to follow Android coding guidelines.  Alternatively, you could try to make a new PR based on this branch, and revert the style changes.  Not sure which is easier.

cupakromer:
Do we still need this PR now that #392 has been merged?

davidgyoung:
These features have been implemented in #392

PR #217, opened 2015-06-24
use a private executor to avoid starving the default queue

davidgyoung:
Simple implementation to avoid maxing out the AsyncTask.THREAD_POOL_EXECUTOR by using it for scan processing threads

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -57,6 +57,8 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 
 /**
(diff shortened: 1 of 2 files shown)


cupakromer:
A grep show that the `ModelSpecificDistanceCalculator` also uses the `THREAD_POOL_EXECUTOR`. If that only runs once I don't think it will be an issue. One concern I have is if you match the number of threads that the `AsyncTask` could also run and this new queue is maxed, do you feel these threads will cause time sharing issues for the CPU with the main app?

davidgyoung:
Yes, the ModelSpecificDistanceCalculator use of a thread is only run once the first time the app runs after install, so it should have negligible impact.

I agree that if a client application is making heavy use of threads then it could conflict with what the BeaconService is doing.   Perhaps it should have a smaller max thread count.  Maybe:

   Runtime.getRuntime().availableProcessors()

Instead of:

   Runtime.getRuntime().availableProcessors()*2+1

In the future the maximum could be configurable or vary depending on whether the app is in the foreground or background.  But I don't think it is worth adding that complexity now until we have a demonstrated compelling need.

PR #218, opened 2015-06-26
Fixes for testCompile issue

gelldur:
http://stackoverflow.com/questions/28729576/cannot-add-a-configuration-with-name-testcompile-as-a-configuration-with-that


davidgyoung:
@gelldur, testing this change, I am unable to run tests without errors.  See the following Gist:

https://gist.github.com/davidgyoung/21acbd3f3e555c29f125

Do you see the same?

gelldur:
Nope. But i don't run test. In few days i will test it

gelldur:
Fixed PR https://github.com/AltBeacon/android-beacon-library/pull/244

PR #235, opened 2015-07-10
Detect various AltBeacon manufacturers in the foreground

davidgyoung:
A recent commit put into 2.3.0 and 2.3.1 inadvertently started using scan filters on Android 5.x devices in the foreground:

https://github.com/AltBeacon/android-beacon-library/commit/66f9698b194dcf389f514d567816477e8da7b47f#diff-cbe9383c235303f12aa05161f0ed907cR177

Normally this would be no problem, but in some cases the scan filters did not match some beacon types, including AltBeacons with manufacturer type codes of 0x00e0 (Google) and 0x004c (Apple).  These unexpected manufacturer codes did not match the filter, causing no detections of these beacons to happen at all.

This change reverts the change so no filters are active in the foreground.  However, beacon transmitters using these non-standard manufacturer codes for AltBeacons will still fail to match the filters in the background, meaning these beacons can only be detected in the foreground.

Beacon buyers beware!

### DIFF
--- a/src/main/java/org/altbeacon/beacon/AltBeaconParser.java
+++ b/src/main/java/org/altbeacon/beacon/AltBeaconParser.java
@@ -41,7 +41,16 @@ public class AltBeaconParser extends BeaconParser {
      */
     public AltBeaconParser() {
         super();
-        mHardwareAssistManufacturers = new int[]{0x0118}; // Radius networks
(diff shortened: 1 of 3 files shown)


davidgyoung:
Change released in 2.3.3

PR #243, opened 2015-07-21
Eddystone url transmitter support

davidgyoung:
Changes to make Eddystone URL transmitter work
* Allow variable length identifiers in transmitted advertisement
* Properly handle additional cases in URL compression/decompression


mharper:
LGTM

PR #244, opened 2015-07-21
Moving to robolectric 3.0

gelldur:
Fixed bug with old robolectric library:
http://stackoverflow.com/questions/28729576/cannot-add-a-configuration-with-name-testcompile-as-a-configuration-with-that

Also i fixed code according to:
https://github.com/robolectric/robolectric/wiki/2.4-to-3.0-Upgrade-Guide

Tests are running and passing in 100%

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,17 +28,15 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.0.0'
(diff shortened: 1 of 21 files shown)


davidgyoung:
Looks good, @gelldur, thanks!

davidgyoung:
Moved to local project branch https://github.com/AltBeacon/android-beacon-library/tree/robolectric_v3.0 to fix one more test before merging

PR #251, opened 2015-07-26
Add path loss distance calculation algorithm

davidgyoung:
Adds experimental support for a new formula for calculating distance based on a path loss formula rather than a curve fitted formula.  This returns better results with low power beacon transmitters and is easier to add calculations for different device models.

This alternate means of calculating distance may be enabled by calling:

```
ModelSpecificDistanceCalculator.setDistanceCalculatorClassName(
        PathLossDistanceCalculator.class.getName());
```

Correction factors for various Android models may be added by doing a 1 meter calibration with that device model, to obtain the average RSSI for that model at 1 meter, then subtracting the known iPhone 5 calibration value at 1 meter.    This yields a "Receiver RSSI Offset", which indicates the relative sensitivity of the Android device's bluetooth receiver compared to an iPhone 5.  For a Nexus 5, this value is -2.5 dB, indicating its receiver is only very slightly less sensitive to Bluetooth signals than an iPhone 5.  For a Samsung Galaxy S5, this value is -21.5 dB, indicating its receiver is much less sensitive to Bluetooth signals than an iPhone 5.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/distance/AndroidModel.java
+++ b/src/main/java/org/altbeacon/beacon/distance/AndroidModel.java
@@ -91,6 +91,56 @@ public class AndroidModel {
         return score;
     }
 
+    /**
+     * Calculates a qualitative match score between two different Android device models for the
+     * purposes of how likely they are to have similar Bluetooth signal level responses
+     * This algorithm takes into account partial matches of model strings, as Samsung devices
+     * commonly have different model name suffixes
+     * @param otherModel
+     * @return match quality, higher numbers are a better match
+     */
+    public double matchScoreWithPartialModel(AndroidModel otherModel) {
+        double score = 0;
+        if (this.mManufacturer.equalsIgnoreCase(otherModel.mManufacturer)) {
+            score = 1;
+        }
+        if (score == 1 ) {
+            score = 1+ratioOfMatchingPrefixCharacters(this.getModel(), otherModel.getModel());
+        }
+        LogManager.d(TAG, "Score is %s for %s compared to %s", score, toString(), otherModel);
+        return score;
+    }
+
+    /**
+     * Returns 1.0 if the string is a complete match, 0.0 if the first characters are different
+     * and 0.5 if the first halves of each string match.  Not case sensitive.
+     * @param string1
+     * @param string2
+     * @return
+     */
+    private double ratioOfMatchingPrefixCharacters(String string1, String string2) {
+        int maxLength = 0;
+        int minLength = 0;
+        int matchingChars = 0;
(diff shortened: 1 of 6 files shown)


HappyDr0id:
Hi David,

You said that for a Nexus 5, the offset is -2.5dB, but on the model-distance-calculations-r2.json file of the path loss branch, you have set -4 on the 'receiver_rssi_offset' field.

Also, you describe that the offset must be supplied by the caller, but what about the slope value ? Is it the same value for every devices ?

Thanks a lot for your work and thanks in advance for your answer ;)

andresteves:
Hi David,

Tried to use the path loss algorithm with no luck, even tweaking the 0.35 constant was of no good. The RSSI values from the json file I used were the ones for Moto G.

The distances came up always less than 1 meter even when the beacons were at 5m and 10m.

I will try and modify the algorithm a little bit to see if it helps but since I am in a tight schedule I might do that afterwards.

Disclose: My company provides bespoke beacon solutions.

davidgyoung:
@andresteves.  Yes, I have also gotten inconsistent results with this algorithm, which is the reason I have not merged this pull request.  It just has not proven to work better (or even as well) as the exiting algorithm in the library.  If you manage to make any changes that improve  things, I would be interested in hearing about the results.

davidgyoung:
iPhone 12 average error: 26%

| Measured| Estimated| Estimate |       |
| Distance| Distance | Error %  |  RSSI |
| ------- | -------- | -------- | ----- |
| 1.0     |  1.0     |  0.0     |   -65 |
| 3.0     |  1.9     |  35      |   -72 |
| 5.0     |  2.8     |  43      |   -76 |

Note: Distances are in meters, RSSI in dBm

davidgyoung:
The iPhone 12 performs particularly poorly using this distance estimation model.  The reason appears to be because the RSSI measurement of a beacon at 1m away is particularly weak relative to the measurements further away at 3m and 5m.  

Why might this be?  I suspect it is because of the placement of the Bluetooth antenna such that RF-blocking components are in behind the antenna on the back side of the phone and attenuate the signal coming from the back side of the phone.  (In the measurements I took, the back side of the phone was facing the beacon).    Any  "shadowing" of the beacon signal by phone components blocking the signal in one direction might be more acute at close range (1m) than greater ranges (3m, 5m) because radio waves would have more of a chance to go around the object blocking the signal.  

If this theory is right, then the iPhone 12 and similar models with antenna placement will perform particularly poorly with distance estimates at short range or using distance estimating techniques based on taking a single device calibration measurement at one meter.

iPhone 12 Bluetooth Antenna

PR #263, opened 2015-08-12
Generalized URL compressor to handle upper-case URL components properly

### LINKED ISSUE #262: Eddystone URL encoder fails to encode shortened URLs containing upper-case letters
The URL `http://goo.gl/C2HC48` fails to match the Eddystone URL encoder regular expression. Converting the entire URL to lower case results in a shortened URL that no longer redirects properly.

mharper:
1. Properly matches upper/mixed-case scheme, e. g. HTTP or HTTPS
2. Properly matches upper/mixed-case domain, e. g. CIRCLE8.NET
3. Properly matches upper/mixed-case WWW
4. Properly matches upper/mixed-case known TLDs, e. g. .COM
5. Properly _preserves_ case in path, e. g. `http://goo.gl/C2HC48`

Fixes #262.


davidgyoung:
Looks good to me

PR #267, opened 2015-08-24
Add callback allowing processing of scanned non-Beacon BLE devices

paulpv:
The AltBeacon library is beautiful (sucking up here a little bit), and fixes a lot of general non-Beacon BLE scanning issues.
If a single app wants to scan both Beacon and non-Beacon BLE devices, then it currently has to spin up two active scanners: 1) AltBeacon library for Beacons, 2) a second android.bluetooth.le.BluetoothLeScanner to scan for non-BLE devices that duplicates all of the AltBeacon's non-Beacon agnostic BLE scanning logic.
This would be both hard to maintain and inefficient use of CPU.

I propose that AltBeason expose a callback that allows processing of non-Beacon BLE devices.
My proposed change is pretty simple and does not harm existing behavior.
An implementation only needs to call "BeaconManager.getInstanceForApplication(...).setNonBeaconScannedCallbacks(...)" to opt in to allowing seeing scanned non-Beacon BLE devices in a "@Override public void onNonBeaconScanned(BluetoothDevice device, int rssi, byte[] scanRecord) { ... }" method.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -38,6 +38,7 @@ import android.os.RemoteException;
 import org.altbeacon.beacon.logging.LogManager;
 import org.altbeacon.beacon.logging.Loggers;
 import org.altbeacon.beacon.service.BeaconService;
+import org.altbeacon.beacon.service.scanner.NonBeaconLeScanCallback;
 import org.altbeacon.beacon.service.RangeState;
 import org.altbeacon.beacon.service.RangedBeacon;
 import org.altbeacon.beacon.service.RunningAverageRssiFilter;
@@ -114,7 +115,7 @@ public class BeaconManager {
     private final ArrayList<Region> monitoredRegions = new ArrayList<Region>();
     private final ArrayList<Region> rangedRegions = new ArrayList<Region>();
     private final ArrayList<BeaconParser> beaconParsers = new ArrayList<BeaconParser>();
-    private BeaconService.NonBeaconScannedCallbacks mNonBeaconScannedCallbacks;
+    private NonBeaconLeScanCallback mNonBeaconLeScanCallback;
     private boolean mBackgroundMode = false;
     private boolean mBackgroundModeUninitialized = true;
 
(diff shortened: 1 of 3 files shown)


davidgyoung:
Thanks for this PR, @paulpv.  I appreciate that this does not make breaking API changes and has a minimal code impact.

I have not come across a use case that needs a feature like this, but I respect that others may have.  Before merging this, I'd love to get input from anybody else with a similar use case to see if they agree this is the best approach.  Another approach might be to always call the callback method regardless of whether a beacon was successfully parsed.  And an option might be to pass a flag indicating whether a beacon was parsed.  Obviously, If the method signature were changed, it would require a new interface for the callback.

Anybody else need raw access to Bluetooth LE scan data for non-beacon devices when using this library?  Any thoughts on this approach?

paulpv:
@bueller? :)

davidgyoung:
I did hear of one other person out-of-band who said he was interested in such a feature.  I'm therefore inclined to merge this unless anyone has an objection.

chrishunterkiller:
i need also this features, thats why in a project i had to extract scanner and change code to get nonBeaconLeDevices, good features imo

davidgyoung:
Thanks, @chrishunterkiller we will be rolling this into the next release

michael-customlbs:
We have the need for exactly this API, but also for iBeacon packets. We need the raw data additionally to the parsed iBeacons.
Is it possible that you call the listener also in case an iBeacon was detected?
For now we forked the project, but obviously this is a pain in the a#%

vitas:
I need it as well.

davidgyoung:
Good news!  This is already available in the 2.7 release.  Binaries is available on JCenter/Maven and available for download here:

https://github.com/AltBeacon/android-beacon-library/releases/tag/2.7

Javadocs are here:

http://altbeacon.github.io/android-beacon-library/javadoc/org/altbeacon/beacon/BeaconManager.html#setNonBeaconLeScanCallback(org.altbeacon.beacon.service.scanner.NonBeaconLeScanCallback)

michael-customlbs:
@davidgyoung: you mean original Pull-request? what about the my proposed change that you can register no mather which type of BLE device?

davidgyoung:
Sorry,  @michael-customlbs, I misunderstood.  

Any thoughts on a backward compatible API change that would allow this variant?  I think the trickiest thing is coming up with how it would be configured via the BeaconManager.  Implementation should be simple.

michael-customlbs:
You are correct, implementation should be simple. Basically move the callback before ``if (beacon != null) {``.

For backwards compatibility is a tough thing here as almost everything contains nonBeacon in the name which makes it difficult to create a generic ``RawBeaconLeScanCallback``.

Issue #173, opened 2015-04-24
Long data fields are not parsed properly


davidgyoung:
Data fields don't parse properly if they are:
 * little endian
 * are longer than four bytes

wns349:
Hello, 
I was wondering if there's any update on this. 

It seems that the `BeaconParser` is still not able to parse data fields correctly.

My custom layout is as follow: (notice data field of size 8)
```
BeaconParser parser = new BeaconParser();
parser.setBeaconLayout("m:2-3=0118,i:4-7,p:8-8,d:9-16,d:18-25");
```

It seems that `BeaconParser` is incorrectly parsing the data string (in hexadecimal format) to Long.

davidgyoung:
There has been no work on this no.  In fact, since I noticed the problem, you are the only person I have heard from who seems to  be interested in the issue.

The data field parsing needs to be re-implemented to populate an arbitrary byte array provide new accessor methods that can get the data fields in this format.

Until then, the data fields are essentially limited to four bytes, and accessors will always return the bytes as as a big-endian signed long integer.

I'd welcome a pull request if you are interested in working on this.  A change should leave the existing accessor methods as-is for backward compatibility, but provide new accessors that provide access to byte arrays.

PR #268, opened 2015-08-25
Parsing large longs instead of crashing on Long.parseLong("0xffffffff")

paulpv:


davidgyoung:
This offers a partial fix to this issue: https://github.com/AltBeacon/android-beacon-library/issues/173

A problem with this solution is that it changes the behavior of the library for identifier fields that are 6 bytes long (e.g. Eddystone-UID instance identifier fields) which are supposed to be converted to a string as hex (e.g. "0x010203040506").  Because this behavior is published, this would essentially be a breaking API change.

 An alternate solution to this would be to not use the string when converting to the long values for the data fields but instead using the underlying byte array.

paulpv:
@davidgyoung Understood.
Would it perhaps be worth adding a flag or parameter to interpret the number as hex or decimal?

davidgyoung:
The Identifier class stores as a byte array internally, and utility methods allow conversion to various formats.   The default string conversion, however needs to remain consistent.

PR #283, opened 2015-09-15
make the region exit time configurable

contrudar:
Hello, I requested the feature that allows to change the exit region time. So, I made the straightforward solution via static variable and methods of BeaconManager where all settings are located.


davidgyoung:
Thanks, @contrudar.  Can you please update the name of the variable to be sExitRegionPeriod?  The Android code standard (not universally followed in this project, but preferred) is to prefix static class variables with "s".

contrudar:
done

davidgyoung:
Looks good to me.

PR #285, opened 2015-09-17
Access static RunningAverageRssiFilter.setSampleExpirationMillisecond…

policybased:
Access static RunningAverageRssiFilter.setSampleExpirationMilliseconds via class reference


davidgyoung:
Nice simplification, thanks!

PR #286, opened 2015-09-17
Access static BeaconManager.getInstanceForApplication via class refer…

policybased:
Access static BeaconManager.getInstanceForApplication via class reference


davidgyoung:
Looks good to me.  Thanks for these changes, @policybased

PR #290, opened 2015-09-22
Use CopyOnWriteArrayList for beacon parser list

matthewmichihara:
BeaconManager used to either return a regular ArrayList or unmodifiableList depending on whether any consumers were bound to the BeaconManager. This unnecessarily caused ConcurrentModificationExceptions if used incorrectly, and was easily bypassed. Replace with the simpler CopyOnWriteArrayList.

I ran the unit tests and did some local testing on a snapshot build and it seemed to be ok.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.2.3'
+        classpath 'com.android.tools.build:gradle:1.3.0'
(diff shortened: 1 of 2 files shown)


davidgyoung:
Looks good.  Thanks, @matthewmichihara

matthewmichihara:
thanks for the quick turnaround. no rush, but any idea when the next release will be cut?

davidgyoung:
We are trying to get a release out by the Android M release date on the 29th.

PR #296, opened 2015-10-04
Only request ACCESS_COARSE_LOCATION permission on M

hdodenhof:
The `ACCESS_COARSE_LOCATION` permission is not necessary on pre-M devices and therefore should only be requested on M.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -5,7 +5,7 @@
     <uses-permission android:name="android.permission.BLUETOOTH"/>
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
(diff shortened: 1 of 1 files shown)


davidgyoung:
Nice.  I didn't even know that XML element existed!

davidgyoung:
@hdodenhof as reported in #521, the Google Play Store has recently started rejecting submissions with both of the following permissions:

``
<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
<uses-permission-sdk-23 android:name="android.permission.ACCESS_COARSE_LOCATION"/>
```

This often happens when two libraries are used, one of which merged in the first permission and the other the second.

What's worse, removing one of these permissions in the parent manifest with an explicit remove fails, apparently due to manifest merger bug.  This prevents folks from submitting apps with this library in some cases, unless the AAR is cracked open to alter or remove the library manifest to remove the conflicting permission.

Unfortunately, I can not think of a better solution than reverting this PR's change and going back to the more general permission.  Yes, we would lose the benefit of not requiring the permission on pre-SDK23 apps. But this is a less bad scenario than folks not being able to submit to the Play Store at all.

Any other ideas before I back this out?

hdodenhof:
@davidgyoung sorry for getting back this late. I just had a quick look and I agree that reverting this change seems to be the best solution for now.

PR #299, opened 2015-10-10
Add proper indentation for code snippet.

danialgoodwin:
In Markdown, for each layer of lists, an additional four spaces need to be added so that code blocks will appear.


davidgyoung:
Thanks!

PR #312, opened 2015-11-05
Fix cancelling wakeup alarm

permutohedron:
As it is stated in the comment when cancelling wakeup alarm it was intended to be defered to the long long future due to some Samsung devices constraints. Actually, the time to deploy alarm parameter was passed with overflow (Long.MAX_VALUE + currentTime() is actually in the past) and the alarm is deployed right after it's creation. Not as it is intended to be. This commit fixes it.


davidgyoung:
Thanks for catching this!

Issue #270, opened 2015-08-27
Switching from monitoring to ranging during active scan cycle breaks scanning


hdodenhof:
Hi David,

thanks for your work on this library! I've been using it for a while but I ran into a small issue lately.

So my app monitors beacons with a specific UUID and as soon as it receives a `didEnterRegion` callback, it stops monitoring and starts ranging that same UUID. The problem is that ranging stops within seconds after it has been started.

The relevant code basically looks like this:

```java
    @Override
    public void didEnterRegion(Region region) {
        mBeaconManager.stopMonitoringBeaconsInRegion(region);
        mBeaconManager.startRangingBeaconsInRegion(region);
    }
```

I did some debugging and these are the steps that lead to the issue:

1. App calls `startMonitoringBeaconsInRegion()`
2. Starting monitoring calls `mCycledScanner.start()` which sets `mScanningEnabled=true` and calls `scanLeDevice(true)`
3. `scanLeDevice(true)` then sets `mScanCyclerStarted=true`, `mScanning=true` and starts a scan cycle
4. App calls `stopMonitoringBeaconsInRegion()`
5. Stopping monitoring calls `mCycledScanner.stop()` which sets `mScanningEnabled=false` and calls `scanLeDevice(false)`
6. `scanLeDevice(false)` then sets `mScanning=false`
7. App calls `startRangingBeaconsInRegion()`
8. Starting ranging calls `mCycledScanner.start()` which sets `mScanningEnabled=true` but does not call `scanLeDevice(true)` as `mScanCyclerStarted` is still `true`
9. At some point in the future `finishScanCycle()` is called as the scan cycle started earlier for monitoring ends
10. `finishScanCycle()` does nothing as `mScanning=false` and `mScanningEnabled=true`

At this point no new scan has been scheduled so scanning stops. It's also impossible to start a new scan as `mScanCyclerStarted=true` and this guards the call to `scanLeDevice(true)` in `start()`. `mScanCyclerStarted` is only reset to `false` in `finishScanCycle()` if `mScanning=true` and `mScanningEnabled=true`, which - if I'm not mistaken - can't be achieved in this situation. Looks like a deadlock.

A workaround for an app is waiting for the scan cycle to finish after calling `stopMonitoringBeaconsInRegion()` and only then call `startRangingBeaconsInRegion()`.

The device I observed this on was a Moto G (2nd gen) with Android 5.0.2 and AndroidLScanning disabled. Library version was 2.5.1.

[...]

PR #317, opened 2015-11-17
Stopping and starting monitoring disables scans

davidgyoung:
This fixes a race condition where stopping monitoring and scanning and then immediately restarting monitoring and scanning leaves the library in a state where it is not doing any scanning at all.  This was caused by a race condition that kept the `mScanCyclerStarted` flag from being set to false if a scan cycle was not allowed to end normally after stopping scanning/monitoring before restarting again.

This change eliminates that race condition by setting the flag immediately.


davidgyoung:
Fixes #270

PR #328, opened 2015-12-21
Add utility to access raw Eddystone-TLM bytes

davidgyoung:
This experimental utility allows recreating the raw bytes of an Eddystone-TLM advertisement.  This is useful, because Google APIs allow reporting of detected telemetry packets to their backend, but want the raw bytes to be sent to the server.  While this library is designed to decode beacon packets into fields based on their layout, and format beacon fields into packets for advertising purposes, it does not easily allow access to raw bytes of a previously detected beacon.  This utility makes that possible.

To use this feature, you will need a copy of the library with this code added.  An ad-hoc release is available for download [here.](https://github.com/AltBeacon/android-beacon-library/releases/tag/ad-hoc-eddystone-telemetry-accessor)

Once your project is compiled with a library version that supports this feature, you can access the raw Eddystone-TLM bytes of any Beacon object with associated Eddystone-TLM fields (stored in the beacon.getExtraDataFields() list) with code like this:

```
byte[] telemetryBytes = new EddystoneTelemetryAccessor().getTelemetryBytes(beaconWithTelemetry);
```

You can also get the raw bytes as a Base64 encoded string with:

```
String encodedTelemetryBytes = new EddystoneTelemetryAccessor().getBase64EncodedTelemetry(beaconWithTelemetry);
```

Example debug output showing this in action:

```
12-26 14:25:58.951 16117-16117/com.radiusnetworks.locate D/EddystoneTLMAccessor: Rehydrated telemetry bytes are :20 00 00 00 4e 23 fc 02 00 00 fc 02 00 00
12-26 14:25:58.951 16117-16117/com.radiusnetworks.locate D/EddystoneTLMAccessor: Base64 telemetry bytes are :IAAAAE4j/AIAAPwCAAA=
```

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconParser.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconParser.java
@@ -661,8 +661,11 @@ public class BeaconParser {
         }
 
         // set power
-        for (int index = this.mPowerStartOffset; index <= this.mPowerEndOffset; index ++) {
-            advertisingBytes[index-2] = (byte) (beacon.getTxPower() >> (8*(index - this.mPowerStartOffset)) & 0xff);
+
+        if (this.mPowerStartOffset != null && this.mPowerEndOffset != null) {
+            for (int index = this.mPowerStartOffset; index <= this.mPowerEndOffset; index ++) {
+                advertisingBytes[index-2] = (byte) (beacon.getTxPower() >> (8*(index - this.mPowerStartOffset)) & 0xff);
+            }
         }
 
         // set data fields
--- a/src/main/java/org/altbeacon/beacon/BeaconParser.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconParser.java
@@ -672,13 +672,13 @@ public class BeaconParser {
         // set data fields
         for (int dataFieldNum = 0; dataFieldNum < this.mDataStartOffsets.size(); dataFieldNum++) {
             long dataField = beacon.getDataFields().get(dataFieldNum);
-
-            for (int index = this.mDataStartOffsets.get(dataFieldNum); index <= this.mDataEndOffsets.get(dataFieldNum); index ++) {
+            int dataFieldLength = this.mDataEndOffsets.get(dataFieldNum) - this.mDataStartOffsets.get(dataFieldNum);
+            for (int index = 0; index <= dataFieldLength; index ++) {
                 int endianCorrectedIndex = index;
-                if (this.mDataLittleEndianFlags.get(dataFieldNum)) {
-                    endianCorrectedIndex = this.mDataEndOffsets.get(dataFieldNum) - index;
+                if (!this.mDataLittleEndianFlags.get(dataFieldNum)) {
+                    endianCorrectedIndex = dataFieldLength-index;
                 }
-                advertisingBytes[endianCorrectedIndex-2] = (byte) (dataField >> (8*(index - this.mDataStartOffsets.get(dataFieldNum))) & 0xff);
+                advertisingBytes[this.mDataStartOffsets.get(dataFieldNum)-2+endianCorrectedIndex] = (byte) (dataField >> (8*index) & 0xff);
             }
         }
         return advertisingBytes;
--- a/src/main/java/org/altbeacon/beacon/BeaconParser.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconParser.java
@@ -449,6 +449,7 @@ public class BeaconParser {
                 if (LogManager.isVerboseLoggingEnabled()) {
                     LogManager.d(TAG, "This is a recognized beacon advertisement -- %s seen",
                             byteArrayToString(typeCodeBytes));
+                    LogManager.d(TAG, "Bytes are: %s", bytesToHex(bytesToProcess));
                 }
             }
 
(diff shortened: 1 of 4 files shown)


fabriziogueli:
David good job, i have tested the utility and it seems work as expected.  It's hard to say if the telemetry data are correctly received from Google server because the backend takes a while to decide it's seen enough data points to be able to start reporting diagnostics. Thank you

davidgyoung:
Glad it works for you, @fabriziogueli.  Please do report back if you can verify Google's servers acceped the data.  

Any sample code for sending the data to the servers would be appreciated, too.

fabriziogueli:
Hi @davidgyoung  and happy new year! Google server generated diagnostics for my beacons, therefore i guess that it accepted the data. I think that you can merge this utility into a formal release of the library. 

I created for a part of my Master Thesis a Wrapper Library using your Beacon Library, Estimote SDK (for beacon's provisioning), NearbyMessages API and i developed a GoogleCloudManager to communicate with Google's Servers easily. Your library was very useful for my thesis, thank you.

davidgyoung:
Thanks for the report @fabriziogueli.  I need to add some test coverage for this, then I will merge.

fabriziogueli:
@davidgyoung  I proceeded in this way for sending data to the server:

- I start monitoring beacons in region
- When i entered a region i start ranging and i stop it after 10 seconds for example

>
```java
@Override
        public void didRangeBeaconsInRegion(Collection<Beacon> beacons, Region region) {
                // Here i add the beacons in a Set to avoid duplicates
        } 

- After 10sec, when the ranging is finished i call the "getForObserved" method of Google Proximity API passing as argument my Set of beacons.

>
```java
 public void getForObserved(Callback callback, Collection<Beacon> beacons, String apiKey) {
    JSONObject  requestBody = new JSONObject();
    JSONArray arrayBeacons = new JSONArray();
    for(Beacon b : beacons)
    {
        JSONObject json = new JSONObject();
            JSONObject advertisedId = new JSONObject()
                        .put("type", //Beacon Type as a String) //for example "EDDYSTONE" or "ALTBEACON"
                        .put("id", Utils.base64Encode(// The beacon identifier, as broadcast by the beacon hardware));
                json.put("advertisedId", advertisedId);
                if( // The type of the beacon is EDDYSTONE)
                {
                    if( //There are Telemetry data){
                        json.put("telemetry", new EddystoneTelemetryAccessor().getBase64EncodedTelemetry(b).trim());                  
                    }
                }
                Timestamp timestamp = new Timestamp(System.currentTimeMillis());
                SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.'000000000Z'");
                format.setTimeZone(TimeZone.getDefault());
                json.put("timestampMs", format.format(timestamp));
                arrayBeacons.put(json);
     }  
    requestBody.put("observations",arrayBeacons);
    JSONArray namespacedType = new JSONArray();
    namespacedType.put(new String("*"));
    requestBody.put("namespacedTypes",namespacedType);
    new AuthTask("beaconinfo:getforobserved?key=" + apiKey, POST/*2*/, requestBody.toString(),  callback).execute();
}



The AuthTask is something like that:

>```java
private class AuthTask extends AsyncTask<Void, Void, Void> {
        public static final String AUTHORIZATION = "Authorization";
        public static final String BEARER = "Bearer ";
        private final String urlPart;
        private final int method;
        private final String json;
        private final Callback callback;
        AuthTask(String urlPart, Callback callback) {
            this(urlPart, GET, "", callback);
        }
        AuthTask(String urlPart, int method, String json, Callback callback) {
            this.urlPart = urlPart;
            this.method = method;
            this.json = json;
            this.callback = callback;
        }
        @Override
        protected Void doInBackground(Void... params) {
            try {
                final String token = GoogleAuthUtil.getToken(ctx, account, SCOPE);
                Request.Builder requestBuilder = new Request.Builder()
                        .header(AUTHORIZATION, BEARER + token)
                        .url(ENDPOINT + urlPart);
                switch (method) {
                    case PUT:
                        requestBuilder.put(RequestBody.create(MEDIA_TYPE_JSON, json));
                        break;
                    case POST:
                        requestBuilder.post(RequestBody.create(MEDIA_TYPE_JSON, json));
                        break;
                    case DELETE:
                        requestBuilder.delete(RequestBody.create(MEDIA_TYPE_JSON, json));
                        break;
                    default: break;
                }
                Request request = requestBuilder.build();
                httpClient.newCall(request).enqueue(new HttpCallback(callback));
            } catch (UserRecoverableAuthException e) {
                // GooglePlayServices.apk is either old, disabled, or not present
                // so we need to show the user some UI in the activity to recover.
                Log.e(TAG, "UserRecoverableAuthException", e);
            } catch (GoogleAuthException e) {
                // Some other type of unrecoverable exception has occurred.
                // Report and log the error as appropriate for your app.
                Log.e(TAG, "GoogleAuthException", e);
            } catch (IOException e) {
                // The fetchToken() method handles Google-specific exceptions,
                // so this indicates something went wrong at a higher level.
                // TIP: Check for network connectivity before starting the AsyncTask.
                Log.e(TAG, "IOException", e);
            }
            return null;
        }
    }

The callback returns any beacon information and attachments accessible to your application as a Json.

I hope I have been clear

davidgyoung:
@fabriziogueli thanks for the sample code.

FYI, I have added a unit test for this, but have discovered the bytes were in reverse order in the generated telemetry due to an endianness bug.  That has been corrected in code now, but I would be surprised if Google accepted your submissions you sent before.

fabriziogueli:
@davidgyoung  it is very strange, maybe Google server generated diagnostics using old data. Three weeks ago i tested NearbyMessages Api that automatically send telemetry data to the server. I will do other tests with the new code and I will let you know

Issue #335, opened 2016-01-29
Using RTC timer breaks periodic scanning when clock is adjusted


lultimouomo:
I have been experiencing a bug on a device without battery backed RTC:
if the device is started without a network connection, and then the network cable is attached later on, beacon ranging stops - i.e., didRangeBeaconsInRegion is never called again.

I think this is due to the fact that at startup the RTC is set in the future (yes it is...), and when the net is connected NTP makes it jump back to the present.
CycledLeScanner uses AlarmManager.RTC_WAKEUP, which probably breaks when the RTC clock is adjusted. Is there a reason for not using AlarmManager.ELAPSED_REALTIME_WAKEUP?

davidgyoung:
The library actually use to to use ELAPSED_REALTIME_WAKEUP, but it was changed to solve a different problem mentioned in #154.  

While the problem you mention is not common, it is possible that it could be easily fixed with a change from:

```
alarmManager.set(AlarmManager.RTC_WAKEUP, System.currentTimeMillis() + milliseconds, getWakeUpOperation());
```

to:

```
alarmManager.set(AlarmManager.ELAPSED_REALTIME_WAKEUP, System.elapsedRealtime() + milliseconds, getWakeUpOperation());
```

Would you be interested in making an testing this change?  If it works, a pull request would be welcome.

lultimouomo:
I can give it a try next week.

davidgyoung:
Closed via #339

PR #339, opened 2016-02-02
Use realtime clock for scanning intervals and timers

lultimouomo:
See issue #335

I branched this off the 2.7.1 tag as the current master does not detect beacons on my device.

Besides using `ELAPSED_REALTIME_WAKEUP` alarms, use `SystemClock.elapsedRealTime()`  whenever checking periodic intervals, to be resilient against clock shifts.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/DetectionTracker.java
+++ b/src/main/java/org/altbeacon/beacon/service/DetectionTracker.java
@@ -1,5 +1,7 @@
 package org.altbeacon.beacon.service;
 
+import android.os.SystemClock;
+
 /**
  * Created by dyoung on 1/10/15.
  */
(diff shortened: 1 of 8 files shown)


davidgyoung:
Thanks for this, @lultimouomo.  At first glance all looks good.  I will do some testing.

Have you verified these changes fix the problem you are seeing on the device without a battery-backed RTD?  Out of curiosity, what is the device?

lultimouomo:
Yes; you can also reproduce the problem on a normal phone by setting the date backward by a day or so while an application using the library is running (note that android studio sorts logcat by date, so it will look like AltBeacon logs stop appearing anyway when you warp back time, but after the fix they are actually just inserted at the front!)

The device is a brandless android set top box, based on an amlogic S802 SOC.

PR #340, opened 2016-02-05
Beacon service multiple bind

permutohedron:
This commit fixes the issue with multiple beacon service consumers binding.
Till now only one cunsumer could be bound due to the fact that only one ServiceConnection existed.
Trying to bind another consumer resulted with no onBeaconServiceConnected() callback as Android does not produce antoher onServiceConnected() callback when same ServiceConnection is used multiple times.


davidgyoung:
Good changes, @permutohedron.  It would be nice to have an automated test for this, but I am not sure how hard that would be with Robolectric.

PR #355, opened 2016-03-16
Adjust the BackgroundPowerSaver so it accepts any context

SnyersK:
A more logical way to use the BackgroundPowerSaver would be to register it yourself.
Assuming `app` is an instance of Application, you can do

`app.registerActivityLifecycleCallbacks(new BackgroundPowerSaver(context));`

As this class is right now, you can already do this.
However, this would mean you register the object twice, which might cause some implications.
I haven't tested what it might do, but registering it twice can't be good.

Also, I'm unsure why you required API 18 for this class, since `Application.registerActivityLifecycleCallbacks` is available since API 14.

This would also make it possible to pass any `Context` in the constructor and still be able to use it.

### NOTE
This is a non-obvious breaking change, since no errors will be thrown if a user updates the library and this change is applied.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/powersave/BackgroundPowerSaver.java
+++ b/src/main/java/org/altbeacon/beacon/powersave/BackgroundPowerSaver.java
@@ -30,17 +30,7 @@ public class BackgroundPowerSaver implements Application.ActivityLifecycleCallba
      *
      */
     public BackgroundPowerSaver(Context context, boolean countActiveActivityStrategy) {
-        if (android.os.Build.VERSION.SDK_INT < 18) {
-            LogManager.w(TAG, "BackgroundPowerSaver requires API 18 or higher.");
-            return;
-        }
-        if (context instanceof Application) {
-            ((Application)context).registerActivityLifecycleCallbacks(this);
-        }
-        else {
-            LogManager.e(TAG, "Context is not an application instance, so we cannot use the BackgroundPowerSaver");
-        }
-        beaconManager = BeaconManager.getInstanceForApplication(context);
+        this(context);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


davidgyoung:
I agree that it is better to have any context be able to be passed in.  This could be accomplished by changing:
 
       if (context instanceof Application) {		
            ((Application)context).registerActivityLifecycleCallbacks(this);		
         }

To

     ((Application)context.getApplicationContext()).registerActivityLifecycleCallbacks(this);		

Without any breaking API changes.  I would prefer not to expose the implementation as using life cycle callbacks or make a breaking API change.

Yes, there is no reason to restrict this to API 18.  The entire library requires API 18 to do anything useful, but not this class specifically.

SnyersK:
I can understand this is an unwanted breaking change.
Though, I am curious about why you do not want to expose the usage of life cycle callbacks.

Is it because it is easier to change to another strategy concerning background detection if all the details are contained in the class?

davidgyoung:
Exactly, yes.  It is possible a different technique may be used in the future, so not exposing it hegges that bet.

Also, for power users like you, there is no reason not to make the user realize it uses the callbacks.  For novice users, though, it is an unnecessary distraction and makes explaining setup more difficult.

SnyersK:
I see. Thank you for explaining it to me.

davidgyoung:
Looks good now, thanks!  Would you mind updating the PR title so it reflects the changes?  Then I think it is ready to merge.

SnyersK:
These are ofcourse 2 commits now, I could still rewrite the history to make it 1 commit if you want?

SnyersK:
I changed it into one commit.

davidgyoung:
Looks great, thanks!

PR #368, opened 2016-04-25
add & remove region in regionboostrap

aristide:
now i can add or remove regions from RegionBootstrap by doing

**

> regionBootstrap = new RegionBootstrap(this);
> Region region = new Region("AnyRegion", null, null, null);
> regionBootstrap.addRegion(region);
> regionBootstrap.removeRegion(region);

**

### DIFF
--- a/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
+++ b/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
@@ -38,6 +38,7 @@ public class RegionBootstrap {
     private List<Region> regions;
     private boolean disabled = false;
     private BeaconConsumer beaconConsumer;
+    private boolean serviceConnected = false;
 
     /**
      * Constructor to bootstrap your Application on an entry/exit from a single region.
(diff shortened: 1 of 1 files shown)


src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -96,6 +97,42 @@ public void disable() {
...
     }
 
+    /**
+     * Add a new region
+     *
+     * @param region
+     */
+    public void addRegion(Region region) {
+        if (!regions.contains(region)) {
+            if (serviceConnected) {
```
davidgyoung: It might be  a good idea to log an error in the else case here -- if the user tries to add a Region when the service is not connected.

aristide: your are right !!!

mmarshak:
Hi

It look like regionBootstrap = new RegionBootstrap(this) is not implemented it, there is no ability to call new RegionBootstrap(this) without specify the region, is there away to start the RegionBoostrap in the inherited class of Appliaction and add later regions?

davidgyoung:
These code changes were moved to a new PR and merged from there.  See: https://github.com/AltBeacon/android-beacon-library/pull/369

mmarshak:
@davidgyoung I looked at 2.9.1 but I did not find a contractor of RegionBootstrap that does not take one argument, all require a region. I think I am missing something?

davidgyoung:
@mmarshak, the pull request discussed above only augmented the class with `addRegion` and `removeRegion` methods.  If you want to construct it without any regions at first, simply pass in an empty (zero length) region list to the existing constructor.

mmarshak:
@davidgyoung , 

Thanks, do you mean the following ?

regionBootstrap = new RegionBootstrap(this);
Region region = new Region();
regionBootstrap.addRegion(region);

or you mean

regionBootstrap = new RegionBootstrap(this);
regionBootstrap.addRegion(null);

davidgyoung:
Try this:

    regionBootstrap =new RegionBootstrap(this, new ArrayList<Region>());

mmarshak:
@davidgyoung Thanks! I will give it a try.

PR #369, opened 2016-04-25
add & remove region from RegionBootstrap 

aristide:


src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -96,6 +97,46 @@ public void disable() {
...
+    public void addRegion(Region region) {
+        if (!regions.contains(region)) {
+            if (serviceConnected) {
+                try {
+                    beaconManager.startMonitoringBeaconsInRegion(region);
+                } catch (RemoteException e) {
+                    LogManager.e(e, TAG, "Can't add bootstrap region");
+                }
+            }else{
+                LogManager.w(TAG, "Adding a rigion: service not yet Connected");
```
davidgyoung: "rigion" should be "region"

davidgyoung:
Looks good, @aristide.  Thank you for the contribution.  This is a nice convenience enhancement.

PR #375, opened 2016-04-29
Missing log arguments due to typo error

javorcd:
Replaced `$` for `%` to display the correct arguments in the log


davidgyoung:
Thanks, @javorcd!

PR #377, opened 2016-05-02
Update BeaconParser.java

mario-pucci:
As already discussed, the library combines the URL schema of advertisement with the real URL but, in the pattern declaration the schema is not present.

We have to increase the url detection from 3-20v to 3-21v in order to accomodate a 18 byte long string (1 for schema and 17 for URL)

MP


davidgyoung:
Thanks, @mario-pucci

mario-pucci:
Thanks David

Issue #143, opened 2015-03-30
getInstanceForApplication should be renamed & is prone to context leaks


samskiter:
getInstanceForApplication or the BeaconManager class should probably call getApplicationContext on the application context to prevent leaks.

The suggested use for the BeaconManager is from an activity, like so:

`private BeaconManager beaconManager = BeaconManager.getInstanceForApplication(this);`

To prevent the activity being leaked, the beaconmanager should really make sure it calls `getApplicationContext` on the passed context.

Secondly `getInstanceForApplication` implies you should be passing an application to the method, simply `getInstance` might be better

davidgyoung:
Fair point.  I think the simplest solution is to change this line:

`client = new BeaconManager(context);`

To:

`client = new BeaconManager(context.getApplicationContext());`

samskiter:
Lgtm. Or put the getApplicationContext() into the constructor.

cupakromer:
Can this be closed now that #381 has been merged?

davidgyoung:
Fixed in #381

PR #381, opened 2016-06-01
Fix 143 potential context leak

ovy9086:
#143 Fix potential memory leak when passing an `Activity` to `BeaconManager.getInstanceForApplication(Context context)` , because the `BeaconManager` would keep a reference to that `Context`.

The fix ensures that `BeaconManager` always keeps a reference to the application `Context`.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -246,7 +246,7 @@ public class BeaconManager {
     }
 
    protected BeaconManager(Context context) {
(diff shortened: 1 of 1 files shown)


davidgyoung:
Thanks, @ovy9086, I have been intending to make this change for some time for other reasons.  Can you please comment on the title of this PR?  What is the specific "leak" issue that you intend to address  with this change.

ovy9086:
@davidgyoung I have edited the initial comment. see above

davidgyoung:
Makes sense.  Thanks!

Issue #385, opened 2016-06-14
Moto X 2nd gen distance coefficients


HappyDr0id:
Hello,

I have done measurements for the Moto X 2nd gen device.

Here are the values :
Meters  RSSI
0.25 	-58
0.5 	        -69
1 		-77
2 		-87
3 		-92
4 		-96
5 		-98
6 		-96
7 		-91
8 		-98
9 		-95
10 		-96
12 		-101
14 		-99
16 		-100
18 		-98
20 		-97
25 		-98
30 		-101

Android device:
version: 6.0
build_number: MPE24.49-18
model: XT1092
manufacturer: Motorola

Beacon Info:
Kontakt.io Tag
Advertisements per second: 10
Transmit Power: Max 4dBm

iPhone 5 Average RSSI @1m: -60

Distance formula coefficients calculated for these values:
 Intercept: -0.45324519
 Multiplier: 0.1862616782
 Power: 8.235367435

Here is the spreadsheet calculations : https://drive.google.com/open?id=19N6ufipDyJMatF8v9iadVG5c6E-0fOEmTZ37wOKuPw0

Would you mind confirm me that calculated values are correct and integrate it to the library please ? Thank you in advance !

HappyDr0id:
Nevermind, if I am not wrong, I understand that the Intercept value is C, the multiplier is A and the power is B.

davidgyoung:
Resolved by #386

PR #386, opened 2016-06-15
Add distance calculations for the Moto X 2nd gen XT1092

HappyDr0id:
Implement feature request #385


davidgyoung:
Thanks, @HappyDr0id.  In your testing, do you see significantly more accurate distance estimates with these coefficients vs. the defaults?

Issue #384, opened 2016-06-13
Better support for custom beacons with extra data advertisement packets (BeaconParser extra layout)


mfatiga:
## Intro

I am developing an app for reacting to custom beacons. The beacon implementation broadcasts multiple different advertisement packets one of which is an identification packet (similar to iBeacon) and the other one is a telemetry packet (similar to Eddystone-TLM). According to [BeaconParser.setBeaconLayout](http://altbeacon.github.io/android-beacon-library/javadoc/org/altbeacon/beacon/BeaconParser.html#setBeaconLayout(java.lang.String)), this library supports parsing custom advertisement packets to extract extra data from them.

## Problem

This does not work in my project because of 2 reasons:

### 1 - no service UUID

`beacon.getServiceUuid()` returns -1 because of which my beacons are automatically dismissed by `GattBeaconTracker.track`, which then doesn't collect any additional telemetry packets for that beacon.

### 2 - advertisement packets don't contain the same byte sequence
The line in [BeaconParser.setBeaconLayout](http://altbeacon.github.io/android-beacon-library/javadoc/org/altbeacon/beacon/BeaconParser.html#setBeaconLayout(java.lang.String)) 
> Signifies that the layout is secondary to a primary layout with the same matching byte sequence (or ServiceUuid)

My beacon's advertisement packets don't contain same matching byte sequences so the library does not relate the 2 beacon parser layouts as being for the same beacon type.

## My current beacon layout example
```
BeaconParser main = new BeaconParser().setBeaconLayout("m:2-3=abcd,i:4-5,i:6-7,i:8-9,p:24-24");
BeaconParser extra = new BeaconParser().setBeaconLayout("x,m:2-2=51,d:21-21");
```
As you can see, the layout's don't have matching services or matching byte sequences.

## Solution proposals

The `beacon.getServiceUuid() != -1` condition in `GattBeaconTracker` could be removed and `GattBeaconTracker` renamed to `BeaconTracker` :-)

### Relate beacon parser beacon layouts by their identifiers
Beacon parser beacon layouts should not be related by matching byte sequences or service UUIDs, they should be related by the `BeaconParser.identifier` field. 

### Change the way extra layouts are handled
The extra data layout should be added separately to the `BeaconParser`, for example:
```
BeaconParser parser = new BeaconParser().setBeaconLayout("m:2-3=abcd,i:4-5,i:6-7,i:8-9,p:24-24");
[...]

PR #387, opened 2016-06-15
Better support for custom beacons with extra data advertisement packets

mfatiga:
Related to issue #384 
Please review and comment with any changes. Thank you!

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -89,7 +89,7 @@ public class BeaconService extends Service {
     private Set<BeaconParser> beaconParsers  = new HashSet<BeaconParser>();
     private CycledLeScanner mCycledScanner;
     private boolean mBackgroundFlag = false;
-    private final GattBeaconTracker mGattBeaconTracker = new GattBeaconTracker();
+    private ExtraDataBeaconTracker mExtraDataBeaconTracker;
     private ExecutorService mExecutor;
 
     /*
@@ -203,12 +203,20 @@ public class BeaconService extends Service {
 
         beaconManager = BeaconManager.getInstanceForApplication(getApplicationContext());
 
+        //flatMap all beacon parsers
+        boolean matchBeaconsByServiceUUID = true;
         if (beaconManager.getBeaconParsers() != null) {
             beaconParsers.addAll(beaconManager.getBeaconParsers());
             for (BeaconParser beaconParser : beaconManager.getBeaconParsers()) {
-                beaconParsers.addAll(beaconParser.getExtraDataParsers());
+                if (beaconParser.getExtraDataParsers().size() > 0) {
+                    matchBeaconsByServiceUUID = false;
+                    beaconParsers.addAll(beaconParser.getExtraDataParsers());
+                }
             }
         }
+
+        //initialize the extra data beacon tracker
+        mExtraDataBeaconTracker = new ExtraDataBeaconTracker(matchBeaconsByServiceUUID);
 
         defaultDistanceCalculator =  new ModelSpecificDistanceCalculator(this, BeaconManager.getDistanceModelUpdateUrl());
         Beacon.setDistanceCalculator(defaultDistanceCalculator);
@@ -405,7 +413,7 @@ public class BeaconService extends Service {
                     "beacon detected : %s", beacon.toString());
         }
 
-        beacon = mGattBeaconTracker.track(beacon);
+        beacon = mExtraDataBeaconTracker.track(beacon);
         // If this is a Gatt beacon that should be ignored, it will be set to null as a result of
         // the above
         if (beacon == null) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/altbeacon/beacon/service/BeaconService.java
```diff
@@ -200,7 +202,14 @@ public void onCreate() {
                 BeaconManager.DEFAULT_FOREGROUND_BETWEEN_SCAN_PERIOD, mBackgroundFlag, mCycledLeScanCallback, bluetoothCrashResolver);
 
         beaconManager = BeaconManager.getInstanceForApplication(getApplicationContext());
-        beaconParsers = beaconManager.getBeaconParsers();
+
+        if (beaconManager.getBeaconParsers() != null) {
+            beaconParsers.addAll(beaconManager.getBeaconParsers());
```
davidgyoung: Does making beaconParsers a `Set` instead of a `List` help here?  Because there is no custom hashCode defined for `BeaconParser` all this will do is make sure that two identical objects are not added to the list.  Is this a concern?

src/main/java/org/altbeacon/beacon/service/GattBeaconTracker.java
```diff
@@ -1,22 +1,18 @@
...
 
 import java.util.HashMap;
 
 /**
  * Keeps track of beacons that have ever been seen and
- * merges them together in the case of Gatt beacons
+ * merges them together depending on configured beacon parsers
  * Created by dyoung on 5/5/15.
  */
 public class GattBeaconTracker {
```
davidgyoung: It might be good to rename this something like `ExtraDataBeaconTracker` since it no longer is specific to GattBeacons

src/main/java/org/altbeacon/beacon/service/GattBeaconTracker.java
```diff
@@ -69,13 +62,12 @@ private Beacon trackGattBeacon(Beacon beacon) {
...
         }
-        mBeacons.put(trackedBeacon.hashCode(), trackedBeacon);
         matchingTrackedBeacons.put(trackedBeacon.hashCode(), trackedBeacon);
+        mBeaconsByKey.put(getBeaconKey(trackedBeacon), matchingTrackedBeacons);
     }
 
-    private String serviceUuidAndMac(Beacon beacon) {
-        return beacon.getBluetoothAddress()+beacon.getServiceUuid();
+    private String getBeaconKey(Beacon beacon) {
+        return beacon.getBluetoothAddress() + beacon.getServiceUuid();
```
davidgyoung: What if the serviceUUID is not -1, but `beacon.isMultiFrameBeacon()` returns true?  Wouldn't this prevent associating extra data between beacons with different serviceUUIDs but the same Mac?  Is this needed in order to ensure backward compatibility?

davidgyoung:
This looks good, @mfatiga.  Do you think you can add test coverage for the new association syntax to `src/test/java/org/altbeacon/beacon/service/GattBeaconTrackerTest.java`?  You can run unit tests on the command line with `./gradlew test`, then see the results in `build/reports/tests/debug/index.html`

src/main/java/org/altbeacon/beacon/service/BeaconService.java
mfatiga: I made it a set to avoid duplicate `BeaconParser`s and optimize the advertisement parsing loop. This is done in case someone adds a `BeaconParser` as an extra data parser and also as a regular parser. But, you are right, adding a custom hashCode would also help.

src/main/java/org/altbeacon/beacon/service/GattBeaconTracker.java
mfatiga: Agreed, I'll do that.

src/main/java/org/altbeacon/beacon/service/GattBeaconTracker.java
mfatiga: I thought about removing the serviceUUID from the "beaconKey" which would make the beacons unique only by their MAC address, but I don't know if this would be ok. What do you think?

mfatiga:
Good to hear :-). I will create a couple of tests for the new association syntax.

src/main/java/org/altbeacon/beacon/service/GattBeaconTracker.java
davidgyoung: If you remove the serviceUuid entirely, it will affect legacy behavior (e.g. it would associate Eddystone-TLM frames with an iBeacon or AltBeacon frame coming from the same hardware device.)  We probably don't want this.  So I don't think it is a good idea to remove it, but you could have it *conditionally* removed if the association were configured programmatically.

mfatiga:
@davidgyoung I have updated the pull request with the proposed changes. I have added tests and tested the library in my project and it works without issues.

src/main/java/org/altbeacon/beacon/service/BeaconService.java
```diff
@@ -203,13 +203,21 @@ public void onCreate() {
 
         beaconManager = BeaconManager.getInstanceForApplication(getApplicationContext());
 
+        //flatMap all beacon parsers
+        boolean matchBeaconsByServiceUUID = true;
```
davidgyoung: This looks good

po5i:
@mfatiga  Could you provide an example of use?

PR #392, opened 2016-06-21
Add Multiple Notifiers Support (Range and Monitor)

opsct:
Re-do the job done by @Maxeee09 in [this pull request](https://github.com/AltBeacon/android-beacon-library/pull/171) with a correct code style


src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -109,9 +109,9 @@
     protected static BeaconManager client = null;
     private final ConcurrentMap<BeaconConsumer, ConsumerInfo> consumers = new ConcurrentHashMap<BeaconConsumer,ConsumerInfo>();
     private Messenger serviceMessenger = null;
-    protected RangeNotifier rangeNotifier = null;
+    protected final List<RangeNotifier> rangeNotifiers = new CopyOnWriteArrayList<>();
```
davidgyoung: I think it would be good if these collections were based on `Set` not `List`.  Because if folks repeatedly call `addXxxNotifier()` with the same object, it could cause trouble.  Making this change could protect against this.  Perhaps use `ConcurrentHashMap.newKeySet()`?

davidgyoung:
@opsct, do you need the `RegionRangeNotifier` and `RegionMonitorNotifier` classes?  I know they were part of the original PR, but they are not referenced by any of the other files, and they appear to me to be utility classes.  They might be useful, but probably won't be unless there is documentation and code examples to go with them.   I'd be inclined to remove them (and `RegionFilter` which is their dependency) unless somebody can make the case for them.

Sorry to make you spin your wheels on this -- I know this is not your original PR.  I appreciate your help.

src/main/java/org/altbeacon/beacon/BeaconManager.java
opsct: I think to use an [CopyOnWriteArraySet](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArraySet.html). What do you think?

As consequences, the getMonitorNotifier() and the getRangeNotifier() method are irrelevants. Do you want to keep them or can I remove them ? 
(I think these methods are only useful internally to the library. They are not in usage any more inside the library)

davidgyoung: We do need to keep `getMonitorNotifier()` and `getRangeNotifier()` so that this is not a breaking API change.  I think a `CopyOnWriteArraySet` is probably fine.

opsct:
The update has been done

PR #401, opened 2016-07-20
Fix Eddystone transmissions to meet spec

davidgyoung:
Users reported that Google's BeaconTools for Android and Google Chrome for Android could not detect Eddystone-URL frames sent by this library.  This turned out to be true, because the transmission did not include the Service UUID in a separate PDU from the Service data.  This change corrects that.

Prior to this change BeaconTools and Chrome for iOS detected Eddystone transmissions from this library just fine.  Android versions of these apps, however, did not.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
@@ -179,15 +179,17 @@ public class BeaconTransmitter {
                         (byte) ((serviceUuid >> 8) & 0xff)};
                 ParcelUuid parcelUuid = parseUuidFrom(serviceUuidBytes);
                 dataBuilder.addServiceData(parcelUuid, advertisingBytes);
-            }
-            else {
+                dataBuilder.addServiceUuid(parcelUuid);
+                dataBuilder.setIncludeTxPowerLevel(false);
+                dataBuilder.setIncludeDeviceName(false);
+
+            } else {
                 dataBuilder.addManufacturerData(manufacturerCode, advertisingBytes);
             }
 
             AdvertiseSettings.Builder settingsBuilder = new AdvertiseSettings.Builder();
 
             settingsBuilder.setAdvertiseMode(mAdvertiseMode);
-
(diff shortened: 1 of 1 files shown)


davidhelms:
This seems like a failure of the beacon tools app and the chrome implementation an Android.

The purpose of the service UUID advertisement is for detection by a mobile client for the purpose of identifying, connecting to and possibly changing the configuration of the beacon. This advertisement should only be present when the unit is in connectable mode which is supposed to be a short lived window of time. Google's Eddystone GATT guidelines recommend having some sort of button or other means to put the beacon in and out of connectable mode.

I'm not sure what the intent of the users' reporting this issue was when they were scanning for use with the beacon tools, but configuration of a mobile device equipped with the ABL does not seem to make sense.

Most likely this seems like it should be reported as a bug to the Google team.

davidhelms:
I realize I misunderstood what was referred to by the service PDU. Disregard my comment

AruLNadhaN:
Still it is not getting detected in Google Chrome(51) - "Physical web pages nearby". I'm using the latest release(2.9).

davidgyoung:
@AruLNadhaN, do you see it detected in the BeaconTools app?  I have tested that I could get detections in both using 2.9, but I know that Physical web pages nearby is very tricky to get to trigger.  BeaconTools is much more reliable.

AruLNadhaN:
@davidgyoung It is getting detected in Beacon Tools app. But it is not getting detected by Chrome :( Do u suggest me any workaround to make it work on the Chrome ?

davidgyoung:
Try launching Chrome on your Android device, hitting the settings menu, then choosing Privacy -> Physical Web and make sure it is turned on.  Make sure you also have Bluetooth and location turned on, and have internet access.  If any one of these things are not set correctly, you won't see anything.

AruLNadhaN:
I have tried every possible methods. Still it is not getting detected.  You can take a look at the transmitter code here - https://gist.github.com/AruLNadhaN/fbbf657bb9141418becf7d2d2e1f74e7

PR #407, opened 2016-07-27
Fix malformed JSON in EidResolver.java

knight7:
- Removes an extra colon that causes an HTTP 400
  Bad Request error when using an API key rather
  than an OAuth token


davidgyoung:
NIce catch, thanks!

Issue #418, opened 2016-08-20
Android N Repeat Region Exit/Entry Events


davidgyoung:
As of Android N, DP4:

> We’ve changed the BLE Scanning behavior starting in DP4. We’ll prevent applications from starting and stopping scans more than 5 times in 30 seconds. For long running scans, we’ll convert them into opportunistic scans.

https://developer.android.com/preview/support.html

With default settings, scanning is stopped and restarted every 1.1 seconds in the foreground (needed for older Bluetooth chipsets and stacks, mostly on devices running older Android version, that only deliver one advertisement callback per device per scan cycle.  Because mostly older devices behave this way, this regular scan restarting probably not needed for devices running Android N.)  

The problem happens on Android N because after 5.5 seconds, (5 scan restarts within 30 seconds) scans get blocked for the next 24.5 seconds, causing a region exit because no beacon packets are detected in that time.  After 30 seconds, a scan succeeds in starting and a new packet detection causes a new entry event.

Proposed solution is to detect if running on Android N, and if so, don't stop scans so frequently, and instead simply continue scanning.

hcknl:
Hello

Seems no trouble at all, works very well on Android N right now  but i wonder that, may continuous scanning cause bluetooth device problem or battery efficiency problem?

davidgyoung:
Thanks for the test.  Continuous scanning does use more battery, which is why it is recommended (and by default) only done by the library when the app is in the foreground.

glose25:
This fixes the issue I was seeing on Nexus 6P with N Preview 5. Thanks.

SteveReeves:
We're having this issue on our 7.0 devices running AltBeacon and I just wanted to clarify what the solution here is.  Do I simply need to change the foreground scan period to something longer (like 6.1 seconds) on Android N devices?

eliaslecomte:
AltBeacon too often stops - starts a BLE scan. By using the latest AltBean, this stop-starting does not happen as frequently on Android N. So using 2.9.1 or later should help you.

SteveReeves:
I updated to 2.9.1 yesterday and I'm still seeing what David mentioned above where I get results returned for the first 5.5 seconds, but the subsequent 24.5 seconds has no beacons showing up in scans.

davidgyoung:
@SteveReeves, two questions:

1. Have you customized your scan rate with the library?
[...]

PR #419, opened 2016-08-20
Avoid spurious entry/exit events on Android N

davidgyoung:
This change limit restarting scans on Android N to once every 6 seconds to comply with the OS-imposed limit of 5 BLE scans every 30 seconds per app.  Without this change, scans at default rates are blocked for 24.5 out of every 30 seconds causing spurious exits and entry events.

Addresses #418

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,8 +49,8 @@ allprojects {
 
 android {
     compileSdkVersion 24
-//    buildToolsVersion "24"
-    buildToolsVersion "23.0.3"
+    buildToolsVersion "24"
+//    buildToolsVersion "23.0.3"
 
     defaultConfig {
         minSdkVersion 7
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,8 @@ allprojects {
 }
 
 android {
-    compileSdkVersion 23
+    compileSdkVersion 24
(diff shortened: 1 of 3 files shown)


davidgyoung:
A pre-release binary for testing is available here:

https://github.com/AltBeacon/android-beacon-library/releases/tag/2.9.1-beta1

To test this with your app on Android N, download the AAR file into a libs folder under your project directory then edit your  build.gradle file to look like this:

```
repositories {
    jcenter()
    flatDir {
        dirs 'libs'
    }
}

dependencies {
    ...
    compile 'org.altbeacon:android-beacon-library:2.9.1-beta1@aar'
}
```

davidgyoung:
Tested this on a Nexus 9 with 7.0 nrd90m.  Verified that with the reference application and library version 2.9 region entry/exit events happen periodically when one beacon is continuously visible.  With this code (2.0.1-beta1) these do not happen.

This is 2.9:

android7-with-2 9

This is 2.9-beta1:

android7-2 9 1-beta

Alfaspider:
Do I need to compile to get aar file, I can't find it on provided link?

davidgyoung:
This is now in the official 2.9.1 release.  No need to compile!

PR #426, opened 2016-09-04
MonitoringState preservation improvement

eliaslecomte:
After a MonitoringStatus restore, beacons that were inside, have their initial lastSeenTime and may trigger quickly as exit - enter. Mark all that are inside again.


davidgyoung:
Looks good to me.

Issue #136, opened 2015-03-02
AltBeacon Library startLeScan of BluetoothAdapter and stopLeScan halts the UI thread.


sanikaj:
I observed every time the StartLeScan of the BluetoothAdapter occurs the UI hangs for a while and so is the same on the StopLeScan. It is similar to the Estimote App issue people were facing. Link of the estimote app is as below:
https://github.com/Estimote/Android-SDK/issues/85. It does it both to the One Plus and the Nexus 5 device. 

OS Version: CyanogenMod 4.4.4 (One Plus)  Nexus 5 (5.0.1)

Devices tested: One Plus,Nexus 5

Snesi:
I have the same issue :worried:

davidgyoung:
Can you please post sample code and steps to reproduce?  This sounds like scanning is happening on the ui thread buyt I have never witnessed this myself.

tomacco:
I have experimented the exact same issue with Cyanogenmod 11.
Checked with Android Device Manager and indeed, bluetooth scanning was running on the UI Thread. I have switched to OmniROM (KitKat as well), and issue has disappeared...

maxsof:
The issue was reproduced on the Motorola Droid Ultra 4.4.4 and Moto E 5.0.2
I guess it can be forced by another bug https://code.google.com/p/android/issues/detail?id=65455
Because when "GKI_exception: out of buffers" exception appears and  beaconManager.startRangingBeaconsInRegion() starts to block the UI thread.

To reproduce "GKI_exception: out of buffers" just keep bluetooth scanning for a night.

lkwjohn:
Have this issue been fixed? I am experiencing similar event that is blocking the main thread (list view) for Android on Nexus 5 as well.

davidgyoung:
I think there may be a number of different issues here, which may result in a similar symptom of laggy UI when scanning for beacons.  Multiple possible sources of a problem include:

* Device or device firmware-specific BLE implementation
* The Android OS version
* The Android Beacon Library version
* Application-specific implementation details

In order to get a good reproducible test case that eliminates application-specific differences, it would be great to get **steps to reproduce using this library's [reference application](https://github.com/altbeacon/android-beacon-library-reference),** and documenting the specific Android version and library version used to reproduce the problem and how it is determined that the UI thread is blocking.

[...]

PR #430, opened 2016-09-14
Start and stop BLE scans from a background thread to prevent blocking the UI (issue #136)

marcosalis:
This pull request fixes the problems reported in issue #136 where ANRs are caused by AltBeacon accessing the BluetoothAdapter (which has heavy internal synchronization and even explicit wait() calls) from the main thread.
A HandlerThread is used for the purpose, and no internal state is accessed from within the posted runnables so that it's not necessary to make the CycledLeScanner classes thread safe. In the same way, callbacks from the BluetoothAdapter are already always executed from the main thread (they're called, internally, from a Handler created with the main looper).

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
+++ b/src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
@@ -12,6 +12,8 @@ import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.os.Build;
 import android.os.Handler;
+import android.os.HandlerThread;
+import android.os.Looper;
 import android.os.SystemClock;
 
 import org.altbeacon.beacon.BeaconManager;
@@ -40,7 +42,10 @@ public abstract class CycledLeScanner {
     private long mScanPeriod;
 
     protected long mBetweenScanPeriod;
-    protected final Handler mHandler = new Handler();
+
+    protected final Handler mHandler = new Handler(Looper.getMainLooper());
+    protected final Handler mScanHandler;
+    private final HandlerThread mScanThread;
 
     protected final BluetoothCrashResolver mBluetoothCrashResolver;
     protected final CycledLeScanCallback mCycledLeScanCallback;
@@ -57,6 +62,10 @@ public abstract class CycledLeScanner {
         mCycledLeScanCallback = cycledLeScanCallback;
         mBluetoothCrashResolver = crashResolver;
         mBackgroundFlag = backgroundFlag;
+
+        mScanThread = new HandlerThread("CycledLeScannerThread");
+        mScanThread.start();
+        mScanHandler = new Handler(mScanThread.getLooper());
     }
 
     public static CycledLeScanner createScanner(Context context, long scanPeriod, long betweenScanPeriod, boolean backgroundFlag, CycledLeScanCallback cycledLeScanCallback, BluetoothCrashResolver crashResolver) {
@@ -154,6 +163,10 @@ public abstract class CycledLeScanner {
             stopScan();
             mLastScanCycleEndTime = SystemClock.elapsedRealtime();
         }
+    }
+
+    public void destroy() {
+        mScanThread.quit();
     }
 
     protected abstract void stopScan();
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -273,6 +273,7 @@ public class BeaconService extends Service {
         LogManager.i(TAG, "onDestroy called.  stopping scanning");
         handler.removeCallbacksAndMessages(null);
         mCycledScanner.stop();
+        mCycledScanner.destroy();
         monitoringStatus.stopStatusPreservation();
     }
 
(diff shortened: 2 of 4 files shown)


davidgyoung:
Thanks, @marcosalis.

One concern I have is that this change appears to enforce that callbacks always happen on the main thread.  This may cause ANR problems for existing applications that do not expect this and upgrade to a version of the library with this change.  This would happen for any app doing heavy processing in a beacon callback.

Thoughts?

marcosalis:
Thank you for your answer @davidgyoung.
Are you referring to the `BluetoothAdapter` callbacks into the AltBeacon library, or the "client code" listeners for ranging/monitoring operations?

`BluetoothAdapter` already delivers all the callback results on the main thread by using a Handler internally, so nothing would change on the interaction with the library (and you don't have to make the `CycledLeScanner` classes thread safe as their state is only accessed from the main thread).

If you're talking about the former, from what I see in the codebase the callbacks are always executed from `BeaconIntentProcessor.onHandleIntent()`, which is always executed from a worker thread (see `IntentService` [documentation](https://developer.android.com/reference/android/app/IntentService.html#onHandleIntent(android.content.Intent))), so nothing should change in the way AltBeacon delivers its results.
I do think, by the way, that the library should deliver the callbacks on the main thread (as I've commented in PR #149), but as you say it's a breaking change so you might want to consider that for future releases ;)

marcosalis:
@davidgyoung any updates on this? Just as additional information and motivation to implement this change, here's a screenshot of a method profiling on the main thread when the `CycledLeScanner` is in action in our application.
As you can notice, the `BluetoothLeScanner` code from the Android SDK is blocking the thread on an `Object.wait()` for around 20 ms, which is not much in absolute terms but it's already above the "16ms" empirical limit in which the main thread should draw a frame on screen.
screen shot 2016-10-13 at 17 52 58

davidgyoung:
Sorry for the delay on this, @marcosalis.  I will do some testing on this today.

src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
```diff
@@ -40,7 +42,10 @@
     private long mScanPeriod;
 
     protected long mBetweenScanPeriod;
-    protected final Handler mHandler = new Handler();
+
+    protected final Handler mHandler = new Handler(Looper.getMainLooper());
+    protected final Handler mScanHandler;
```
davidgyoung: Is there a reason to use a `HandlerThread` here?  Why not just construct `mScanHandler = new Handler();`?  Then you don't need to clean up the handler thread later.

davidgyoung:
I've tested this code, and it works for me, but I have some questions about the handler setup I'd like to resolve before merging.

src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
```diff
@@ -40,7 +42,10 @@
     private long mScanPeriod;
 
     protected long mBetweenScanPeriod;
-    protected final Handler mHandler = new Handler();
+
+    protected final Handler mHandler = new Handler(Looper.getMainLooper());
```
marcosalis: The default, parameterless `Handler` constructor uses the `Looper` from the current thread (as documented [here](https://developer.android.com/reference/android/os/Handler.html#Handler()) and [here](https://developer.android.com/training/multiple-threads/communicate-ui.html)).

In this particular case, `CycledLeScanner` is always instantiated from the main thread (from `BeaconService.onCreate()`) and consequently, that `Handler` will already use the main thread's looper.
This addition doesn't modify the current logic, it's only a futureproof guarantee that the class behaviour won't change if in the future if `CycledLeScanner` is instantiated from another thread that doesn't have a `Looper` (you would immediately notice because the Handler creation would crash).

src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
marcosalis: @davidgyoung see my answer below (and this [doc](https://developer.android.com/reference/android/os/Handler.html#Handler())), this is the change that makes the whole PR work: by creating it with `new Handler()` you'd be using the `Looper` from the thread that's instantiating `CycledLeScanner`, which is the main thread.

Instead, by creating a `HandlerThread` (which is just a thread with a Looper attached to it) we make sure the "expensive" Bluetooth code from the SDK is posted and executed in our own worker thread.

davidgyoung: OK, got it now... apologies, it's been awhile, it's been awhile since I looked at this code.  I was incorrectly thinking that new `Handler()`  always creates a new thread.

davidgyoung:
Thanks for your patience here, @marcosalis.  Another question:  Wouldn't it be simpler to create a single `Handler` on a new thread, and run all of the code in CycledLeScanner on this thread?  That way we wouldn't need both `mHandler` and `mScanHandler` and the work could be run on the one handler from the` CycledLeScanner#start` and` CycledLeScanner#stop` methods.

marcosalis:
@davidgyoung you could use this approach, but then you'd have to make all the `CycledLeScanner` (and concrete implementations) internal state thread-safe.
At the moment, the internal state is only accessed by one thread (the main thread), so you don't have to complicate your class state and add explicit synchronisation. The `mScanHandler` is now only used to start/stop the `BluetoothAdapter` scans and doesn't access any instance field from its class.

davidgyoung:
Fair point.  Ok, looks good to me.

marcosalis:
Thank you @davidgyoung !

Issue #429, opened 2016-09-09
support bluetooth permissions with attribute required="false"


giorgio-zamparelli:
### Expected behavior
The library should work with the following permissions in the manifest.xml (note the android:required="false")
<uses-permission android:name="android.permission.BLUETOOTH" **android:required="false"** />
<uses-permission android:name="android.permission.BLUETOOTH_ADMIN" **android:required="false"** />

The required="false" otherwise the app will not be installed on devices that do not have bluetooth.
In many apps the bluetooth functionality is a nice plus but should not forbid devices without bluetooth to use the app anyway without the bluetooth features

### Actual behavior
The app crashes at startup with error "android required false Need BLUETOOTH permission: Neither user 10077 nor current process has android.permission.BLUETOOTH"

### Steps to reproduce this behavior
Add the android:required="false" after the BLUETOOTH and BLUETOOTH_ADMIN permissions in the manifest

### Mobile device model and OS version
Samsung Galaxy S6 Edge SM-G925F
Android 6.0.1

### Android Beacon Library version
2.9.1

davidgyoung:
Good point, thanks for the report.  I know this was tested with earlier Android versions so I suspect that this crash may be new to Android 6.

chyrta:
Had the same issue on Samsung device came from my Rollbar stats. Definitely needs to be fixed.

davidgyoung:
@giorgio-zamparelli and @chyrta, do either of you have a test device you can use to reproduce this to see if my proposed fix in #438 solves the problem?

Also, @giorgio-zamparelli in your initial report, you say this happened with a Samsung Galaxy S6 Edge SM-G925F.  But that device has Bluetooth LE.  How is it possible that you saw the crash on that device?

giorgio-zamparelli:
yes I do own Samsung Galaxy S6 Edge SM-G925F and I will try to test whether #438 fixed the problem.
Not sure how but the crash is caused by the android:required="false" attribute if though the device has both Bluetooth and BLE.

davidgyoung:
I suspect this crash may be caused by Samsung Knox, a security framework which allows disabling bluetooth permissions for apps installed inside a Knox container.  The solution to this problem probably the same as the solution to that problem. 

See here:  https://github.com/AltBeacon/android-beacon-library/issues/440

PR #438, opened 2016-10-02
Fix crash on startup for devices without BLE 

davidgyoung:
This is an attempt to fix crashes reported in #429, where devices without BLE crash on startup. 

This does a check to see if the BLUETOOTH_ADMIN permission is granted before trying to access the BluetoothAdapter, which is what causes the crash in the bug report.  If the permission is not granted, the subsequent operations are not performed.

I have tested that this does not break existing functionality on devices with BLE, but I have not tested that this actually prevents a crash on devices without BLE, because I have no such test device.  If anybody has such a device and can test this, please let me know.  I am reluctant to merge this change if we cannot test first that it does something beneficial.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
@@ -49,6 +49,11 @@ public class BeaconTransmitter {
      * @param parser specifies the format of the beacon transmission
      */
     public BeaconTransmitter(Context context, BeaconParser parser) {
+        if (context.checkCallingOrSelfPermission("android.permission.BLUETOOTH_ADMIN") !=
+                PackageManager.PERMISSION_GRANTED) {
+            LogManager.e(TAG, "Application does not have BLUETOOTH_ADMIN permission");
+            return;
+        }
         mBeaconParser = parser;
         BluetoothManager bluetoothManager =
                 (BluetoothManager) context.getSystemService(Context.BLUETOOTH_SERVICE);
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -32,6 +32,7 @@ import android.bluetooth.BluetoothDevice;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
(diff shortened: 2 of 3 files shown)


cupakromer:
I think this may cause more problems than it fixes. Based on where the checks are there are a lot of internal fields to the classes which are not initialized. This will cause numerous possible `NullPointerException` issues downstream.

This library has always had an implicit requirement that Bluetooth permissions are available. As this lib is meant to be a relatively low-level API I'm not sure this type of requirement is any different than the underlying Android OS permissions required for calling methods on Android's [`BluetoothAdapter`](https://developer.android.com/reference/android/bluetooth/BluetoothAdapter.html). Requiring bluetooth permissions is at a different level of requirement than requiring BLE support. Throwing a `SecurityException` is generally the proper course of action to take when an library/API requires a permission and it has not been granted.

As the manifest already attempts to declare these permissions, for an app to actively override them means they are trying to run in an environment where the API may not properly be supported. Based on how Android handles such situations, I think documenting that the lib/classes require `BLUETOOTH`/`BLUETOOTH_ADMIN` permissions and leaving it up to the app developer to verify this before calling into the API is the way to go. 

Aside from adding this to the Java docs and README we could use the Android [`@RequiresPermission`](https://developer.android.com/studio/write/annotations.html#permissions) to further help app developers guard against this.

davidgyoung:
Fair, points, @cupakromer. 

I think the key issue is that the absent permissions are an unexpected side effect of bluetooth not being present on some devices.  Because folks typically don't have a non-bluetooth capable device with which to test, they don't find out about these exceptions until their apps are in the wild.  This is why I was trying to handle it silently, but I agree about the null pointer issues, and I am nervous about not being able to test this.

Maybe a better solution is to explicitly attempt to get a BluetoothAdapter when the BeaconManager is constructed (and let the operating system throw the` SecurityException`).  This way, the exception won't be thrown at a later time when it won't be caught.  App developers who want to enable their apps to run on devices without bluetooth can then be instructed to adjust the bluetooth permissions to required="false" and catch the SecurityException when obtaining  a `BeaconManager`.

It's still going to be very hard for folks to properly test this case in their apps, because they won't be able to exercise that code path without a physical device that doesn't have Bluetooth.

davidgyoung:
Closed in favor of #444

Issue #414, opened 2016-08-16
java.lang.IllegalStateException in android.bluetooth.le.BluetoothLeUtils.checkAdapterStateOn


catchmartin:
### Expected behavior

Exception to be caught by library, in event of unusual IllegalStateException
### Actual behavior

Exception not caught and causes app crash.

```
java.lang.IllegalStateException: BT Adapter is not turned ON
at android.bluetooth.le.BluetoothLeUtils.checkAdapterStateOn(BluetoothLeUtils.java:136)
at android.bluetooth.le.BluetoothLeScanner$1.handleMessage(BluetoothLeScanner.java:85)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:168)
at android.app.ActivityThread.main(ActivityThread.java:5885)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:797)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:687)
```
### Steps to reproduce this behavior

Exact sequence not found but appears to be happening predominantly on HTC devices on app start. This has been reported from two apps we are involved with. Additionally, found the same issue reported on http://stackoverflow.com/q/37965687/5580666
### Mobile device model and OS version

Description from one of the app teams:

"It seems that there is some problem with some new HTC phones (like htc one_e8 dual sim) that are running latest Android 6.0. Possibly this is bug in their firmware. This bug is on the 6th place with total crashes of 147.

It probably causes app to crash on start, so It's not possible for users to use the app when bluetooth is turned on. It's reported even in Google Play."
### Android Beacon Library version

Seen in version 2.6.1 in the two apps we are involved with and also reported in version 2.7 on http://stackoverflow.com/q/37965687/5580666

IMPORTANT:  This forum is reserved for feature requests or reproducible bugs with the library itself.  If you need help with using the library with your project, please open a new question on StackOverflow.com.

AlexLardschneider:
The same issue happens in my app.

It seems to only happen on HTC devices running 6.0.1 and as soon as the user disables bluetooth while the app is scanning for beacons. The beacon handler in my app gets unbound as soon as bluetooth is disabled (i.e. the bluetooth receiver is called), but it still crashes.

The issue happens with AltBeacon 2.7 - 2.9.

davidgyoung:
[...]

Issue #306, opened 2015-10-20
Getting ClassNotFoundException org.altbeacon.beacon.service.StartMRData


ankitgupta100:
With the latest Android 6 update (Nexus 5), I am receiving following error on debugging - 

E/Parcel: Class not found when unmarshalling: org.altbeacon.beacon.service.StartRMData
E/Parcel: java.lang.ClassNotFoundException: org.altbeacon.beacon.service.StartRMData
E/Parcel:     at java.lang.Class.classForName(Native Method)
E/Parcel:     at java.lang.Class.forName(Class.java:324)
E/Parcel:     at android.os.Parcel.readParcelableCreator(Parcel.java:2383)
E/Parcel:     at android.os.Parcel.readParcelable(Parcel.java:2337)
E/Parcel:     at android.os.Message.readFromParcel(Message.java:571)
E/Parcel:     at android.os.Message.-wrap0(Message.java)
E/Parcel:     at android.os.Message$1.createFromParcel(Message.java:527)
E/Parcel:     at android.os.Message$1.createFromParcel(Message.java:525)
E/Parcel:     at android.os.IMessenger$Stub.onTransact(IMessenger.java:51)
E/Parcel:     at android.os.Binder.execTransact(Binder.java:453)
E/Parcel:  Caused by: java.lang.ClassNotFoundException: org.altbeacon.beacon.service.StartRMData
E/Parcel:     at java.lang.Class.classForName(Native Method)
E/Parcel:     at java.lang.BootClassLoader.findClass(ClassLoader.java:781)
E/Parcel:     at java.lang.BootClassLoader.loadClass(ClassLoader.java:841)
E/Parcel:     at java.lang.ClassLoader.loadClass(ClassLoader.java:469)
E/Parcel:     at java.lang.Class.classForName(Native Method) 
E/Parcel:     at java.lang.Class.forName(Class.java:324) 
E/Parcel:     at android.os.Parcel.readParcelableCreator(Parcel.java:2383) 
E/Parcel:     at android.os.Parcel.readParcelable(Parcel.java:2337) 
E/Parcel:     at android.os.Message.readFromParcel(Message.java:571) 
E/Parcel:     at android.os.Message.-wrap0(Message.java) 
E/Parcel:     at android.os.Message$1.createFromParcel(Message.java:527) 
E/Parcel:     at android.os.Message$1.createFromParcel(Message.java:525) 
[...]

PR #450, opened 2016-11-09
CycledLeScanner : a version scanning when screen turns on or off

opsct:
This PR permits to create new CycledScanner strategies and permits to get faster beacon detection when doing long scans (more than 10s scan).

It comes with a new CycledScanner strategy, the CycledScannerScreenState class, that scans for beacons only when screen turns on or turn off. 

We split the CycledLeScanner class in two classes:
* LeScanner class that is doing the scanning job (with a LeScannerForJellyBeanMr2 and one LeScannerForLollipop)
* The CycledLeScanner that manages the calls to the LeScanner class

We introduce the notion of "ScanPeriods" and "ActiveScanPeriods".

The scanPeriods is simple object that groups the scanPeriod and betweenScanPeriod paremeter.    
It helps solving the problem of the minimal scanning period on Android L and help us to manage two "ScanPeriods" the "normal" and the "active"

The "ActiveScanPeriods" is usefull when an application is in background, doing long scans (more than 20s) with long delays between the scans.   
The "ActiveScanPeriods" is the "ScanPeriods" used when scanning. Said like this It could sound strange but we notice that we get better scanning result (and faster beacons detection) if an application scans 1s or 6s during 30s than when scanning one time for 30s.
So, the "ActiveScanPeriods" is the "ScanPeriods" that permits to scan 1s or 6s during 30s (as example).

We will be glad to get your feedbacks,

Regards,

Olivier

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,11 +1,11 @@
-###Expected behavior
+### Expected behavior
 
-###Actual behavior
+### Actual behavior
 
-###Steps to reproduce this behavior
+### Steps to reproduce this behavior
 
-###Mobile device model and OS version
+### Mobile device model and OS version
 
-###Android Beacon Library version
+### Android Beacon Library version
 
 IMPORTANT:  This forum is reserved for feature requests or reproducible bugs with the library itself.  If you need help with using the library with your project, please open a new question on StackOverflow.com.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Android Beacon Library
 =======================
 
-[![Build Status](https://snap-ci.com/AltBeacon/android-beacon-library/branch/master/build_image)](https://snap-ci.com/AltBeacon/android-beacon-library/branch/master)
+[![Build Status](https://circleci.com/gh/AltBeacon/android-beacon-library.png?circle-token=4e11fb0dccaa8b98bc67fdbe38b179e4a7d07c27)](https://circleci.com/gh/AltBeacon/android-beacon-library)
 
 An Android library providing APIs to interact with beacons.  Please visit the
 [project website](http://altbeacon.github.io/android-beacon-library/) for how to use this library.
--- a/README.md
+++ b/README.md
@@ -58,13 +58,13 @@ replacing `${altbeacon.version}` with the version you wish to use.
 ## How to build this Library
 
 This project uses an AndroidStudio/gradle build system and is known working with Android Studio
-1.0.1 and Gradle 2.2.1
+2.1 and Gradle 2.2.1
 
 Key Gradle build targets:
 
     ./gradlew test # run unit tests
     ./gradlew build # development build
-    ./gradlew release  # release build  
+    ./gradlew release -Prelease # release build
     ./gradlew generatereleaseJavadoc -Prelease
 
 ## License
--- a/build.gradle
+++ b/build.gradle
@@ -28,14 +28,12 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
-        classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.12.0'
+        classpath 'com.android.tools.build:gradle:1.5.0'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.0'
         classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:3.0.3'
     }
 }
 
-apply plugin: 'android-sdk-manager'
 apply plugin: 'com.android.library'
 
 apply from: 'gradle/eclipse.gradle'
@@ -50,8 +48,8 @@ allprojects {
 }
 
 android {
-    compileSdkVersion 23
-    buildToolsVersion "23"
+    compileSdkVersion 24
+    buildToolsVersion "24.0.2"
 
     defaultConfig {
         minSdkVersion 7
@@ -59,6 +57,7 @@ android {
         versionCode 1
         versionName version
         testInstrumentationRunner "com.google.android.apps.common.testing.testrunner.GoogleInstrumentationTestRunner"
+        consumerProguardFiles 'proguard-rules.pro'
     }
 
     compileOptions {
@@ -117,8 +116,6 @@ dependencies {
         exclude module: 'wagon-provider-api'
         exclude group: 'com.android.support', module: 'support-v4'
     }
-
-    testCompile 'com.android.support:support-v4:21.0.2'
 
     testCompile 'org.mockito:mockito-core:1.10.19'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ ext {
  * Gets the version name from the latest Git tag
  */
 def getVersionName = {
-    def stdout = new ByteArrayOutputStream()
+    /*def stdout = new ByteArrayOutputStream()
     try {
         exec {
             commandLine 'git', 'describe', '--tags'
@@ -19,7 +19,8 @@ def getVersionName = {
     catch (e) {
         println("Can't get version from git: " + e);
         return "adhoc"
-    }
+    }*/
+    return "2.9.1-CS6"
 }
 
 buildscript {
@@ -31,12 +32,18 @@ buildscript {
         classpath 'com.android.tools.build:gradle:1.5.0'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.0'
         classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:3.0.3'
+        classpath "net.researchgate:gradle-release:2.0.2"
+        classpath "com.github.dcendents:android-maven-plugin:1.2"
     }
 }
 
 apply plugin: 'com.android.library'
-
+apply plugin: 'net.researchgate.release'
+apply plugin: 'com.github.dcendents.android-maven'
 apply from: 'gradle/eclipse.gradle'
+
+archivesBaseName = 'android-beacon-library'
+group = 'org.altbeacon'
 
 allprojects {
     version = "${getVersionName()}${isSnapshot == true ? "-SNAPSHOT" : ""}"
@@ -66,11 +73,11 @@ android {
         targetCompatibility JavaVersion.VERSION_1_7
     }
 
-    sourceSets {
-        androidTest {
-            setRoot('src/test')
-        }
-    }
+    /* sourceSets {
+         androidTest {
+             setRoot('src/test')
+         }
+     }*/
 
     lintOptions {
         abortOnError false
@@ -143,9 +150,9 @@ task distribution(dependsOn: [bundleEclipse, build, clean, renameAarForRelease])
     println "Building with version=$version"
 }
 
-task release(dependsOn: 'distribution') << {
+/*task release(dependsOn: 'distribution') << {
     println('Doing release build')
-}
+}*/
 
 android.libraryVariants.all { variant ->
 
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ def getVersionName = {
         println("Can't get version from git: " + e);
         return "adhoc"
     }*/
-    return "2.9.1-CS6"
+    return "2.9.1-CS7"
 }
 
 buildscript {
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ ext {
  * Gets the version name from the latest Git tag
  */
 def getVersionName = {
-    /*def stdout = new ByteArrayOutputStream()
+    def stdout = new ByteArrayOutputStream()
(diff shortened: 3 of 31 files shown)


davidgyoung:
Thanks for the PR, @opsct.  These are interesting ideas, and the addition of extra scanning on screen change is a good one.

Any chance you can provide more details on the test conditions that led you to conclude "we get better scanning result (and faster beacons detection) if an application scans 1s or 6s during 30s than when scanning one time for 30s"?  I would be curious to know the device model/OS version used to see this, the beacon transmission rate, and the difference in the detection time when using this technique.

opsct:
We observe on all the devices we test from Android 4.3 to Android 7 that when a device scans for a long period (for exemple 30s.) the result of scan is delivered at the end of the scan period when using the AltBeacon library.
(If the device just starts to scan for 30s. when the beacon starts to emit, the device will detect the beacon after ~ 28s, but if the device already scan since 15s, the device will detect the beacon after ~ 14s)
To test we simply set the foregroundScanningPeriod to 30000:

     beaconManager.setForegroundScanPeriod(30000);

--------------------------------------
We did additional test today. We think the problem can come from the AltBeacon library delivery process. 
We build a very simple app to detect beacons using directly the scanner.startScan()

    ScanSettings settings = new  ScanSettings.Builder().setScanMode(ScanSettings.SCAN_MODE_LOW_LATENCY).build();
        scanner.startScan(null, settings, scanCallback);

With this app, devices detect entry and exit of our beacons immediately when scanning for 30s. 
We will do further investigations in the next days.

davidgyoung:
@opsct, I believe this is working as designed.  The  detected beacons are generally delivered at the end of the scan cycle, except when the scan cycle is long and the detection causes a new entry condition, in which case they are delivered immediately to avoid a long detection delay.

opsct:
Ok I see and looking deeper to the code I get a better understanding of the delivery process (the staff done in the "onCycleEnd" method from the CycledLeScanCallback class and the "processFromScan" method).

So with these informations, the "ActiveScanPeriods" seems useless and we will remove it. Do you agree?

davidgyoung:
@opsct, yes, this may explain your observation that "we notice that we get better scanning result (and faster beacons detection) if an application scans 1s or 6s during 30s than when scanning one time for 30s."  This would happen  if beacons are already detected recently, so results would not get delivered until the end of the scan cycle.

davidgyoung:
Very good PR, thank you!    I think this is a very good feature that is important to add.  Aside from the comments above, my only reluctance to merge right away is out of concern over this creating hard to track down bugs.  There are lots of code changes here — over 1000 lines, and this is probably the most complex part of the library.  I am not fully confident that after reviewing it that the changes haven’t created little problems.  Unfortunately, unit test coverage of this section of code is pretty much non-existent.  (Although splitting out the LeScanner from the CycledLeScanner would make adding test coverage easier by being able to have a mock LeScanner for unit tests.)

A few general questions and comments:

1. How much testing have you done with this, and under what configurations?

2. Can you update the PR description to remove the discussion of the active cycle, since this has been removed from the PR?

3. Can you please update the branch to include the latest changes from master?

4. Is there any reason for this not to be the default implementation of CycledLeScanner?  The only effective change is for it to give an extra scan period on screen illumination.  I think library users would typically want this, it would still be possible for folks who do not want it for some reason, and changing the default implementation does not affect the public API in any way.

opsct:
1. I did tests on 4 devices (Samsung S, S5, S6 and Nexus 5 plus)
I will redo tests when finishing to merge. Additional tests are welcome.

2. Done

3. In Progress. I have got a question about the "isBluetoothOn()" method in "CycledLeScannerForLollipop" class .
Is that normal that the "isBluetoothOn()" method is implemented only for "CycledLeScannerForLollipop" and not for "CycledLeScannerForJellyBeanMr2" ?

4. The "CycledLeScannerScreenState" is only scanning when the screen goes on or off.  
If I get your suggestion is to permit to the default implementation to  scan in addition of the current scanning period when the screen goes on or off, is it?
I can code this in a next PR.

opsct:
1. I did tests on 4 devices (Samsung S, S5, S6 and Nexus 5 plus)
I will redo tests when finishing to merge. Additional tests are welcome.

2. Done

3. In Progress. I have got a question about the "isBluetoothOn()" method in "CycledLeScannerForLollipop" class .
Is that normal that the "isBluetoothOn()" method is implemented only for "CycledLeScannerForLollipop" and not for "CycledLeScannerForJellyBeanMr2" ?

4. The "CycledLeScannerScreenState" is only scanning when the screen goes on or off.  
If I get your suggestion is to permit to the default implementation to  scan in addition of the current scanning period when the screen goes on or off, is it?
I can code this in a next PR.

davidgyoung:
@opsct, to answer your question about `isBluetoothOn()`, the reason a method is in `CycledLeScannerForLollipop` and not in the MR2 version is because it exists in order to protect against a crash found with Android 6 ROMs.  Because the crash did not happen with Android 4.x ROMs, checking for bluetooth on before starting/stopping scanning was not necessary.

Read here for more info:  https://github.com/AltBeacon/android-beacon-library/issues/414

opsct:
Merge done.

I did tests on Samsung S3 - Android 4.4, S4 - Android 5.0.1, S5 - Android 6.0.1, and Nexus 5 plus - Android 7.0

It seems ok  for me.

I let you cross check.

davidgyoung:
Thanks, @opsct, I'll take a look.

davidgyoung:
@opsct, In testing today with the reference app on my Nexus 5X, I am seeing that background scanning continues continually after putting the app to the background.   I would expect a 30 second duty cycle on scans.  See log below, and note that the scan cycle doesn't stop after putting the app to the background.  You always see "Waiting to stop scan cycle for another 1000 milliseconds".   I don't think this is right.

Do you see this, too?

```
12-11 13:17:01.428 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Scan started
12-11 13:17:02.083 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Set scan periods called with 10000, 300000 Background mode must have changed.
12-11 13:17:02.083 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: We are in the background.  Setting wakeup alarm
12-11 13:17:02.126 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Set a wakeup alarm to go off in 300000 ms: PendingIntent{adb82de: android.os.BinderProxy@70a8cbf}
12-11 13:17:02.461 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Waiting to stop scan cycle for another 1000 milliseconds
12-11 13:17:02.474 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Set a wakeup alarm to go off in 300000 ms: PendingIntent{adb82de: android.os.BinderProxy@70a8cbf}
12-11 13:17:03.511 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Waiting to stop scan cycle for another 1000 milliseconds
12-11 13:17:03.514 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Set a wakeup alarm to go off in 300000 ms: PendingIntent{adb82de: android.os.BinderProxy@70a8cbf}
12-11 13:17:04.556 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Waiting to stop scan cycle for another 1000 milliseconds
12-11 13:17:04.561 17077-17077/org.altbeacon.beaconreference D/CycledLeScanner: Set a wakeup alarm to go off in 300000 ms: PendingIntent{adb82de: android.os.BinderProxy@70a8cbf}
```

opsct:
I think you are not on the last version.

This problem was solved in the [commit](https://github.com/AltBeacon/android-beacon-library/pull/450/commits/5889cb4c0a72915282d71be66d2e3f410bcf5116)

davidgyoung:
@opsct, I'm sorry this has been left open for so long.  I really like this feature and would like to see it part of the library.

Two issues have come up since this was submitted that prevent this from being merged as-is:

1.  Adding the  `getCycledLeScanner` and `setCycledLeScanner` methods on `BeaconManager` is a problem, because in cases where the beacon scanner is in a different process per #306 there will be multiple instances of this class per app, and will cause problems with changes in #479.  We need a different way of communicating this to the `CycledLeScanner`, perhaps by making `ScreenStateBroadcastReceiver` local to the CycledLeScanner.

2. As of Android "O", a long running scan service will not be possible, and an alternate means of scanning will be needed per #484.  In this case, the `ScreenStateBroadcastReceiver` will need to immediately start a scan another way.  But if we can come up with an alternative to the first item, I think we can apply to this one.

opsct:
Hello David,

I am back,
I take time to look to your PR #484 

1. About the getCycledLeScanner/setCycledLeScanner methods, a possibility is to use reflexion.
The getCycledLeScanner/setCycledLeScanner methods can be replaced by the a getCycledLeScannerClass/setCycledLeScannerClass methods.     
The CycledLeScannerClass will be used to initialize the CycledLeScanner in the beaconService or next in the ScanHelper class, using reflexion.    
What do you think about ?

2. The ideal would be to get a JobService launched on "SCREEN_ON" or "SCREEN_OFF", but it's not a possibility that offers the JobScheduler of Android O.    
So, the only possibility I see is to continue to use the ScreenStateBroadcastReceiver, but registered/unregistered in the ScanJob class.

davidgyoung:
Hi, @opsct.

Yes, it would be possible to trigger an immediate ScanJob through a message received by a  broadcast receiver.  This would be a relatively minor change to the `scheduled-job-scanning` branch.  

As to your other point, "About the getCycledLeScanner/setCycledLeScanner methods, a possibility is to use reflexion...."  Is this even necessary?  I'd really like to see screen-on scanning be automatic without having to configure a special version of CycledLeScanner.  It doesn't seem there is a good reason *not* to do this.

opsct:
One interest I see to use JobScheduler on version prior to Android O is to permit to the library to run on devices which already kill background services, (mostly when the screen turns off).

It's the cases on 2 Wiko devices and 2 Huawai devices on which I recently test the library.
(As default, there is a battery saver program which is killing all background services when the screen goes off. The user can decide to autorise an app to run in background but it's a complicated settings that most user won't do)

To get the beaconService runs in background on these devices, I use a JobScheduler which is monitoring the beaconService at regular interval, and restart it when the beaconService has been killed.

Note: 
* the AlarmManager is not working on these devices
* The 4 devices are under Android 6

opsct:
Hi David,

I get the point that you would like the "Screen Turns Off / Screen Turns On" scanning feature integrated as default to the CycledLeScanner.

I would like to get the possibility to configure the CycledLeScanner to develop different background scanning strategies.
I think to strategies based on beacon density: if I summarize, the background scan is higher when it's an area with high beacon density with a user is moving than when it's an area with few beacons or no beacons at all. 
It's something we start to develop as a hack of the background scanning period. We can integrate it to the library in a next step if you think it's interesting. 

About the way to register the "Screen On/ Screen Off" action on Android O: I don't think to launch a JobScheduler when BroadcastReceiver receives a screen action.

The broadcast screen action must be registered and unregistered manually.
And so, I imagine: 
* to register the SCREEN actions (Intent.ACTION_SCREEN_OFF and ACTION_SCREEN_ON) when the ScanJob starts and to unregister it when the ScanJob finishes
* to keep the ScreenStateBroadcastReceiver calling the CycledLeScanner
* To change the way the CycledLeScanner instance is get to avoid using the BeaconManager instance.

Whatever, I can start to integrate the Screen Turns Off / On to the default CycledLeScanner to permit a merge. 
Once merged, I can update the code into the Android 0 branch.

davidgyoung:
When the `ScanJob` is used, I think the `Intent.ACTION_SCREEN_OFF` and `ACTION_SCREEN_ON` can be registered **at all times** and the intents can point at the existing `StartupBroadcastReceiver`.  When the event is received it can simply kick off an extra immediate `ScanJob`.  In the scheduled-job-scanning-android-o branch, this is exactly what happens when an `Intent` is received based on a background bluetooth detection.   So you can simply add a call for that case as well, which should be a pretty simple change.  See the line of code here:

https://github.com/AltBeacon/android-beacon-library/blob/scheduled-job-scanning-android-o/src/main/java/org/altbeacon/beacon/startup/StartupBroadcastReceiver.java#L43

The only difference in the screen on/off case is that the `scanResults` List would be empty.

As to the other point "I think to strategies based on beacon density: if I summarize, the background scan is higher when it's an area with high beacon density with a user is moving than when it's an area with few beacons or no beacons at all."  We should probably discuss this in a new Github issue.  Although it is related, I would like to keep the change separate if possible.  I'd also like to understand more why higher beacon densities would benefit from a different scanning strategy -- it isn't obvious to me immediately why this is true.

opsct:
I get your point. I haven't noticed the immediate scan job.

Ok to discuss later on a new issue about the possibility of other scanning strategies.

paulpv:
Close as fixed by #941 ?

Issue #469, opened 2017-01-11
NPE in BackgroundPowerSaver onActivityPaused and onActivityResume


jeffkelsey:
### Expected behavior
No crash reports

### Actual behavior
Top 2 crashers in our app based on Fabric data is:

#1 
BackgroundPowerSaver line 66
org.altbeacon.beacon.powersave.BackgroundPowerSaver.onActivityResumed

Fatal Exception: java.lang.RuntimeException: Unable to resume activity {com.ticketmaster.mobile.android.na/com.ticketmaster.mobile.android.library.ui.activity.EventDetailsAndFindTicketsCombinedActivity}: java.lang.NullPointerException
       at android.app.ActivityThread.performResumeActivity(ActivityThread.java:2958)
       at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:2987)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2380)
       at android.app.ActivityThread.access$900(ActivityThread.java:164)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1268)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:157)
       at android.app.ActivityThread.main(ActivityThread.java:5377)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:515)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1265)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1081)
       at dalvik.system.NativeStart.main(NativeStart.java)

Crash #2 
BackgroundPowerSaver line 76
org.altbeacon.beacon.powersave.BackgroundPowerSaver.onActivityPaused

Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'void org.altbeacon.beacon.BeaconManager.setBackgroundMode(boolean)' on a null object reference
       at org.altbeacon.beacon.powersave.BackgroundPowerSaver.onActivityPaused(SourceFile:76)
       at android.app.Application.dispatchActivityPaused(Application.java:247)
       at android.app.Activity.onPause(Activity.java:1652)
       at android.support.v4.app.FragmentActivity.onPause(SourceFile:440)
       at com.ticketmaster.mobile.android.library.activity.DrawerActivity.onPause(SourceFile:223)
[...]

PR #475, opened 2017-03-02
Fix power saver crash

davidgyoung:
This is an attempt to fix crashes in the BackgroundPowerSaver reported in #469 caused by a null BeaconManager.  This change moves the construction of the BeaconManager before the registration of activity lifecycle callbacks, so if an exception is thrown creating the BeaconManager, it will not be left in a null state.

It is unclear whether this will solve the problem, so this small change is being put in a test build.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/powersave/BackgroundPowerSaver.java
+++ b/src/main/java/org/altbeacon/beacon/powersave/BackgroundPowerSaver.java
@@ -44,8 +44,8 @@ public class BackgroundPowerSaver implements Application.ActivityLifecycleCallba
             LogManager.w(TAG, "BackgroundPowerSaver requires API 18 or higher.");
             return;
         }
-        ((Application)context.getApplicationContext()).registerActivityLifecycleCallbacks(this);
         beaconManager = BeaconManager.getInstanceForApplication(context);
+        ((Application)context.getApplicationContext()).registerActivityLifecycleCallbacks(this);
     }
 
     @Override


cupakromer:
I wonder if it's possible that once you register the activity lifecycle callback if it's possible for another thread to call this instance before the constructor returns. If that's possible, your re-ordering would definitely fix it.

I wonder if we should also mark `beaconManager` as `final` to indicate that once it's set it won't change.

cupakromer:
I was able to reproduce. I was wrong about how this could happen before 😬. It is related to threading, it just can easily happen when everything runs in a single threaded manner (>ლ).

Here's a new test file which reproduces the problem and shows that your fix works! 👍 

```java
// src/test/java/org/altbeacon/beacon/powersave/BackgroundPowerSaverTest.java
package org.altbeacon.beacon.powersave;

import android.annotation.TargetApi;
import android.app.Activity;
import android.app.Application;
import android.content.Context;
import android.os.Build;

import org.altbeacon.beacon.BeaconManager;
import org.junit.Before;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.robolectric.Robolectric;
import org.robolectric.RobolectricTestRunner;
import org.robolectric.annotation.Config;

import static org.hamcrest.Matchers.is;
import static org.hamcrest.Matchers.nullValue;
import static org.junit.Assert.assertThat;

@Config(sdk = 18)
@RunWith(RobolectricTestRunner.class)
public class BackgroundPowerSaverTest {
    @Before
    public void before() {
        BeaconManager.setsManifestCheckingDisabled(true);
    }

    @Test
    public void initializingHandlesSingleThreadedCallOrder() throws Exception {
        final Activity anyActivity = Robolectric.buildActivity(Activity.class).create().get();
        final Context crashApp = new Application() {
            @Override
            public Context getApplicationContext() {
                return this;
            }

            @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
            @Override
            public void registerActivityLifecycleCallbacks(ActivityLifecycleCallbacks callback) {
                callback.onActivityPaused(anyActivity);
            }
        };

        Exception thrown = null;
        try {
            new BackgroundPowerSaver(crashApp);
        } catch (Exception e) {
            thrown = e;
        }
        assertThat(thrown, is(nullValue()));
    }
}
```

davidgyoung:
Right, that test demonstrates that this PR does fix one case -- I'm just not confident that it is the same case reported by the user.  I was hoping to get feedback with confirmation that it solves that specific problem.

That said, there is not much point in waiting for that feedback before merging since we know it does solve one problem.  I'll simply leave that issue open for now in the hope that we still get that confirmation.

Issue #472, opened 2017-02-09
StackOverflowError


tekinalper:
### Expected behavior
No crash
### Actual behavior
Fatal Exception: java.lang.StackOverflowError
       at android.os.Parcel.writeString(Parcel.java:516)
       at org.altbeacon.beacon.Region.writeToParcel(Region.java:262)
       at android.os.Parcel.writeParcelable(Parcel.java:1292)
       at org.altbeacon.beacon.service.RangingData.writeToParcel(RangingData.java:63)
       at android.os.Parcel.writeParcelable(Parcel.java:1292)
       at android.os.Parcel.writeValue(Parcel.java:1211)
       at android.os.Parcel.writeArrayMapInternal(Parcel.java:618)
       at android.os.Bundle.writeToParcel(Bundle.java:1692)
       at android.os.Parcel.writeBundle(Parcel.java:643)
       at android.content.Intent.writeToParcel(Intent.java:7251)
       at android.app.ActivityManagerProxy.startService(ActivityManagerNative.java:3024)
       at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1566)
       at android.app.ContextImpl.startService(ContextImpl.java:1548)
       at android.content.ContextWrapper.startService(ContextWrapper.java:494)
       at org.altbeacon.beacon.service.Callback.call(Callback.java:69)
       at org.altbeacon.beacon.service.BeaconService.processRangeData(BeaconService.java:402)
       at org.altbeacon.beacon.service.BeaconService.access$300(BeaconService.java:78)
       at org.altbeacon.beacon.service.BeaconService$1.onCycleEnd(BeaconService.java:364)
       at org.altbeacon.beacon.service.scanner.CycledLeScanner.finishScanCycle(CycledLeScanner.java:251)
       at org.altbeacon.beacon.service.scanner.CycledLeScanner.scheduleScanCycleStop(CycledLeScanner.java:243)
       at org.altbeacon.beacon.service.scanner.CycledLeScanner.scanLeDevice(CycledLeScanner.java:215)
       at org.altbeacon.beacon.service.scanner.CycledLeScanner.finishScanCycle(CycledLeScanner.java:284)
...
       at 
       at org.altbeacon.beacon.service.scanner.CycledLeScanner$1.run(CycledLeScanner.java:239)
       at android.os.Handler.handleCallback(Handler.java:733)
       at android.os.Handler.dispatchMessage(Handler.java:95)
[...]

PR #476, opened 2017-03-02
Fix stackoverflow crash when starting/stopping scanning

davidgyoung:
This attempts to prevent a stackoverflow crash from #472 seen in crash reports by checking first to see if Bluetooth LE is supported on the device before starting or stopping scanning.   It is unclear whether this will fix the problem, but it is being put into a PR and an ad-hoc build so the solution can be tested.

Binary release is here: https://github.com/AltBeacon/android-beacon-library/releases/tag/fix-scan-stackoverflow-crash

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
+++ b/src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
@@ -17,6 +17,7 @@ import android.os.Looper;
 import android.os.SystemClock;
 
 import org.altbeacon.beacon.BeaconManager;
+import org.altbeacon.beacon.BleNotAvailableException;
 import org.altbeacon.beacon.logging.LogManager;
 import org.altbeacon.beacon.startup.StartupBroadcastReceiver;
 import org.altbeacon.bluetooth.BluetoothCrashResolver;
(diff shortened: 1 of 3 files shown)


src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
```diff
@@ -169,6 +170,13 @@ public void destroy() {
         mScanThread.quit();
     }
 
+    protected boolean canScan() {
+        if (!mContext.getPackageManager().hasSystemFeature(PackageManager.FEATURE_BLUETOOTH_LE)) {
+            return false;
+        }
+        return true;
+    }
```
cupakromer: This can be simplified to just:

```java
protected boolean canScan() {
  return mContext.getPackageManager().hasSystemFeature(PackageManager.FEATURE_BLUETOOTH_LE);
}
```

cupakromer:
I'm a little concerned that add this so deep inside the call chain will setup some odd state leading to strange behavior. Is there a reason why we need it so far into the method calls?

davidgyoung:
I think this is the wrong solution based on subsequent feedback in #472, so I am going to close this an open a new solution.

davidgyoung:
Closed in favor of #492

PR #482, opened 2017-03-24
Format/code cleanup

cupakromer:
Quick summary:

- Minor markdown formatting update to `CONTRIBUTING.md` so headers render in Github
- Remove legacy `@TargetApi` annotations for Android APIs less than the current min: 7
- Minor lint cleanup of `Parcel3recycle` in tests
- Clarify method behavior of internal helper `BeaconParser#byteArraysMatch`

  This removes potential ambiguity with variable size arrays:

  ```java
  byte[] a1 = { 1, 0, 1, 0 };
  byte[] a2 = { 1, 0, 1 };
  byteArraysMatch(a1, 0, a2, 0);
  byteArraysMatch(a1, 2, a2, 0);
  byteArraysMatch(a2, 1, a1, 1);
  ```

  In the context of parsing BLE payloads we already know which array is the "definitive bytes". Additionally, our usage is always to compare the entire array. Leaving such ambiguity opens up the potential for incorrectly matched payloads. Issues caused by this will be costly to troubleshoot as they will likely only occur far downstream.

  Since this is not a generic utility method this chooses a more restrictive, but contextually explicit, method signature:

  ```java
  boolean byteArraysMatch(byte[] source, int offset, byte[] expected)
  ```

  The same example from before now produces:

  ```java
  byte[] a1 = { 1, 0, 1, 0 };
  byte[] a2 = { 1, 0, 1 };
  byteArraysMatch(a1, 0, a2);    // true
  byteArraysMatch(a1, 2, a2);    // false
  byteArraysMatch(a2, 1, a1);    // false
  ```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,30 +1,30 @@
-#How to Contribute to the Android Beacon Library
+# How to Contribute to the Android Beacon Library
 
-This project welcomes code contributions from the community.  Proposed code changes should be submitted as a pull request on Github.  Please follow the following guidelines when submitting a [pull request](https://github.com/altbeacon/android-beacon-library/pulls).
+This project welcomes code contributions from the communityProposed code changes should be submitted as a pull request on Github. Please follow the following guidelines when submitting a [pull request](https://github.com/altbeacon/android-beacon-library/pulls).
 
-##Style
+## Style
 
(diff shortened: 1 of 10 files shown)


davidgyoung:
Looks good to me.

Thanks for the cleanup, @cupakromer

Issue #473, opened 2017-02-09
Missed beacon observations on 6.0.1 and 5.1 devices


MadisonBlake:
### Expected behavior

Consistent iBeacon observations ranged

### Actual behavior

When scanning on a 6.0.1 and 5.1 device at 10Hz with beacons running at 10Hz observations are frequently missed, sometimes for multiple seconds.

This behavior is not observed on 7.1 device (Pixel).

The same code was ran on three devices gathering time and strengths of beacon range entries.
The following is a table summarizing the time difference (in ms) between successive observations 

|             | μ   | σ   | L∞   |
|-------------|-----|-----|-------|
| Droid Turbo | 301 | 633 | 10547 |
| MotoE       | 525 | 483 | 4922  |
| Pixel     | 173 | 79  | 560   |

Further when I test with my own beacon parser (using `BluetoothLeScanner.StartScan`), I get behavior consistent with 7.1 device on the 6.0.1 device.
To me this indicates it isn't hardware issue.
Maybe I am missing a configuration step somewhere?

### Steps to reproduce this behavior

Have an iBeacon within range
set foreground scan period to be 100ms
set foreground between scan period to be 0ms
record beacon as they come into the RangeNotifier object

If needed I can send code to stimulate this issue.

### Mobile device model and OS version

Encountered on:  
- Droid Turbo 6.0.1
- Moto E 5.1

### Android Beacon Library version

2.9.2

IMPORTANT:  This forum is reserved for feature requests or reproducible bugs with the library itself.  If you need help with using the library with your project, please open a new question on StackOverflow.com.

MadisonBlake:
I did some investigating and it seems to be a problem with Android when starting/stopping the scanner frequently. 

Is there a reason range updates are only done at the end of a scan?
I think it would be generally useful to either allow ranging updates to happen on a frequency differently than scanning or have ranging updates happen on any received beacon.

I made some modifications to support that. I will probably drop those changes in a fork sometime soon. I doubt you will want to blindly pull my changes in, but it may be a good to see what I was thinking as useful.

davidgyoung:
Android  7 did add a restriction that prevents starting and stopping the scanner more than 5 times in a 30 second period.  There was a fix put in the library (which is included in 2.9.2) to address this here:

https://github.com/AltBeacon/android-beacon-library/commit/7bfc78e044f7ea7db3b8514bcbcf51f4071524a7

[...]

PR #491, opened 2017-04-08
Don't restart scanning if device detects duplicate advertisements per scan 

davidgyoung:
**Background:** 

Restarting scanning (stopping and immediately restarting) is necessary on many older Android devices like the Nexus 4 and Moto G because once they detect a distinct BLE packet in a scan, subsequent detections do not get a scan callback.  Stopping scanning and restarting clears this out, allowing subsequent detection of identical advertisements.

On most newer devices, however, this is not necessary, and multiple callbacks are given for identical packets detected in a single scan.   Because the library has no simple way of knowing whether the device can detect duplicate advertisements per scan, scanning is restarted regardless.

**Change**

This adds a detection of duplicate identical advertisements in a single scan.   If duplicates are ever detected, scans are no longer restarted.

**Benefits**

The primary benefit of this change is that **there will no longer be brief dropouts in detections when scans are stopped and restarted**, causing advertisement packets and detections to be missed as described in #473.  This change will allow **much shorter scan intervals while keeping detections reliable**.  A secondary benefit may be some power savings with some bluetooth chipsets, although this savings has not been confirmed or characterized.

**Typical Impact on New Devices**

The flag indicating whether or not a device supports multiple detections per scan is not persisted, so it is effectively cleared on restart.  On devices supporting multiple detections per scan, restarting scanning will still happen.  For default library settings and beacons transmitting at 10 Hz, this determination will happen within a second or two, causing restarts to quickly cease.  

**Typical Impact on Old Devices**

Older devices not supporting multiple detections per scan will see no change, aside from some extra processing done to track distinct packets in each scan cycle.   This extra processing is worth the tradeoff, especially as the number of such devices shrinks.

Text on Android 7.0 with beacon turned on at  04-08 14:43

```
$ adb logcat | grep stopping

04-08 14:42:55.881 16746 16746 D CycledLeScanner: stopping bluetooth le scan
04-08 14:42:56.999 16746 16746 D CycledLeScanner: Not stopping scan because this is Android N and we keep scanning for a minimum of 6 seconds at a time. We will stop in 4882 millisconds.
04-08 14:42:58.102 16746 16746 D CycledLeScanner: Not stopping scan because this is Android N and we keep scanning for a minimum of 6 seconds at a time. We will stop in 3779 millisconds.
04-08 14:42:59.206 16746 16746 D CycledLeScanner: Not stopping scan because this is Android N and we keep scanning for a minimum of 6 seconds at a time. We will stop in 2675 millisconds.
04-08 14:43:00.313 16746 16746 D CycledLeScanner: Not stopping scan because this is Android N and we keep scanning for a minimum of 6 seconds at a time. We will stop in 1569 millisconds.
04-08 14:43:01.471 16746 16746 D CycledLeScanner: Not stopping scan because this is Android N and we keep scanning for a minimum of 6 seconds at a time. We will stop in 410 millisconds.
04-08 14:43:02.472 16746 16746 D CycledLeScanner: Not stopping scanning.  Device capable of multiple indistinct detections per scan.
04-08 14:43:03.585 16746 16746 D CycledLeScanner: Not stopping scanning.  Device capable of multiple indistinct detections per scan.
```

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -50,6 +50,7 @@ import org.altbeacon.beacon.distance.ModelSpecificDistanceCalculator;
 import org.altbeacon.beacon.logging.LogManager;
 import org.altbeacon.beacon.service.scanner.CycledLeScanCallback;
 import org.altbeacon.beacon.service.scanner.CycledLeScanner;
+import org.altbeacon.beacon.service.scanner.DistinctPacketDetector;
 import org.altbeacon.beacon.service.scanner.NonBeaconLeScanCallback;
 import org.altbeacon.beacon.startup.StartupBroadcastReceiver;
 import org.altbeacon.bluetooth.BluetoothCrashResolver;
@@ -90,7 +91,8 @@ public class BeaconService extends Service {
     private boolean mBackgroundFlag = false;
     private ExtraDataBeaconTracker mExtraDataBeaconTracker;
     private ExecutorService mExecutor;
-
+    private final DistinctPacketDetector mDistinctPacketDetector = new DistinctPacketDetector();
+    
     /*
      * The scan period is how long we wait between restarting the BLE advertisement scans
      * Each time we restart we only see the unique advertisements once (e.g. unique beacons)
@@ -359,6 +361,7 @@ public class BeaconService extends Service {
 
         @Override
         public void onCycleEnd() {
+            mDistinctPacketDetector.clearDetections();
             monitoringStatus.updateNewlyOutside();
             processRangeData();
             // If we want to use simulated scanning data, do it here.  This is used for testing in an emulator
@@ -479,6 +482,13 @@ public class BeaconService extends Service {
             }
             if (beacon != null) {
                 mDetectionTracker.recordDetection();
+                if (!mCycledScanner.getDistinctPacketsDetectedPerScan()) {
+                    if (!mDistinctPacketDetector.isPacketDistinct(scanData.device.getAddress(),
+                            scanData.scanRecord)) {
+                        LogManager.i(TAG, "Non-distinct packets detected in a single scan.  Restarting scans unecessary.");
+                        mCycledScanner.setDistinctPacketsDetectedPerScan(true);
+                    }
+                }
                 trackedBeaconsPacketCount++;
                 processBeaconFromScan(beacon);
             } else {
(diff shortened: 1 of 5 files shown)


src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
```diff
@@ -165,6 +165,14 @@ public void stop() {
         }
     }
 
+    public boolean getDistinctPacketsDetectedPerScan() {
+        return mDistinctPacketsDetectedPerScan;
+    }
+
+    public void setDistinctPacketsDetectedPerScan(boolean detected) {
+        mDistinctPacketsDetectedPerScan = detected;
+    }
```
cupakromer: Can we drop these accessors?

This is an internal flag that's only needed for this class. Exposing it means long term support for this flag and limits the ability of this class to be refactored / modified without breaking API changes.

src/main/java/org/altbeacon/beacon/service/BeaconService.java
```diff
@@ -90,7 +91,8 @@
     private boolean mBackgroundFlag = false;
     private ExtraDataBeaconTracker mExtraDataBeaconTracker;
     private ExecutorService mExecutor;
-
+    private final DistinctPacketDetector mDistinctPacketDetector = new DistinctPacketDetector();
```
cupakromer: Is there a reason to put this in the service and not the scanner?

davidgyoung: This needs to be on the service because the service class is where an 
AsyncTask is kicked to do scan processing on another thread.  This processing needs to be done there and not on the main thread as it is it is non-trivial.   This also causes the accessors to be needed on CycledLeScanner

cupakromer:
I want to say I do like the idea of this feature 👍 

Having just recently gone through the cycler/scanner code I found it difficult to reason about the expected behavior given the complexity of the interactions. Currently `CycledLeScanner` has 6 flags and 5 time trackers. The `CycledLeScannerForLollipop` subclass adds an additional 2 flags and 2 time trackers. On top of that logic for when scans start/restart and how that happens is spread across both classes as well (for example Android M. Then there is also the `CycledLeScannerForJellyBeanMr2` which doesn't add any new flags/timers but also needs to be considered in the logic flow to understand how various Android devices will behave.

I'm concerned adding this 7th flag, which is managed by another external class, will make things more difficult to reason through.

I guess it would help to understand how much of `BeaconService`, `CycledLeScanner`, `CycledLeScannerForJellyBeanMr2`, and `CycledLeScannerForLollipop` you consider part of the "public" API locked into semver and how much you consider "internal implementation details". I also ask this in light of what further changes #484 and #450 will require.

davidgyoung:
In general, classes in the service package are not intended to be part of the public API, although this is not explicitly documented anywhere.  As a result, **I'm not super concerned about maintenance overhead of adding public methods to these classes.**  There really isn't any useful way to use these public methods when using the library in binary form -- you really have to modify the source code to do this.

There are some folks who have done exactly this by building modified versions of the CycledLEScanner as in #450, but again, these changes require other modifications to non-public methods in source code in order to make use of them.

The number of flags unfortunately is a reflection of the complexity here.  This is partly in order to accommodate a number of different Android OS versions and device-specific behaviors.  That said, I think this additional flag is worth the additional complexity given the compelling benefit it provides.

PR #492, opened 2017-04-08
Don't start scanning service if BLE unsupported

davidgyoung:
If BLE is unsupported on a device, there is no point in starting the scanning service.  The library has long refused to start the scanning service if on a pre-Android 18 device (which does not support BLE APIs).  So the effects of not starting the scanning service at startup under this similar condition are well known and already in the wild.

This may prevent a stackoverflow crash reported in  #472, which was believed to happen on a device with Android 4.3+ but no BLE support.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -296,6 +296,10 @@ public class BeaconManager {
             LogManager.w(TAG, "Not supported prior to API 18.  Method invocation will be ignored");
             return;
         }
+        if (!mContext.getPackageManager().hasSystemFeature(PackageManager.FEATURE_BLUETOOTH_LE)) {
+            LogManager.w(TAG, "This device does not support bluetooth LE.  Will not start beacon scanning.");
+            return;
+        }
         synchronized (consumers) {
             ConsumerInfo newConsumerInfo = new ConsumerInfo();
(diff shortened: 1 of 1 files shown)


davidgyoung:
Actually, the existing checks are needed anyplace but in `bind()`, as if the service is not bound, those other methods will not do anything.  However, removing the existing checks would cause a change in behavior that a `RemoteException` would be thrown from those other methods if the service had not previously been bound. The checks essentially suppress this Exception and cause it to fail silently.   As a result, removing those checks could cause new crashes for improper API usage.

Unfortunately, this probably does mean performing both the API 18 and BLE available check everywhere.

cupakromer:
I'm not sure I understand your comment. My suggestion _was_ to put both checks everywhere. I just suggested putting them in a single internal helper method `isBleAvailable`. That way there is a single place to add additional checks, if necessary, in the future without having to hunt down every possible place in the code to add such a check.

davidgyoung:
Changes implemented as described.   Testing referenceapp on an emulator without BLE logs:

```
04-17 12:28:59.810 3332-3332/org.altbeacon.beaconreference W/BeaconManager: This device does not support bluetooth LE.
04-17 12:28:59.810 3332-3332/org.altbeacon.beaconreference W/BeaconManager: Method invocation will be ignored.
```

Testing on Nexus 5X with bluetooth on detects beacons as expected.

PR #494, opened 2017-04-09
Ensure singleton object creation thread safety

cupakromer:
This fixes possible race conditions creating the singleton instances of:

- `BeaconManager`
- `DetectionTracker`
- `MonitoringStatus`
- `Stats`

Both `DetectionTracker` and `Stats` require no argument to their constructors. The singleton instances are also both `private` to the class. This means we can create the singleton instance as a constant without worry about any additional threading issues; Java guarantees constant creation thread safety (see _Effective Java_, Second Edition, _Item 3: Enforce the singleton property with a private constructor or an enum type_).

However, both `BeaconManager` and `MonitoringStatus` require the application context in order to be created. This means we cannot use a constant since we won't have the application context instance available at constant creation time. Instead this uses the lazy-loading double check locking pattern with a private lock object. While this double check locking idiom may seem a bit odd it is a well published and tested pattern. The source of the idiom is cited in the local code comments. We also change the instances to be `volatile` to ensure that any thread that reads the field will see the most recently written value - which is necessary since there is no locking when we have already initialized the field. For more details on this pattern see _Effective Java_, Second Edition, _Item 71: Use lazy initialization judiciously_.

This additionally fixes a possible denial-of-service condition with the previous `MonitoringStatus` singleton implementation. This DOS issue can occur because the synchronization is performed by locking the externally accessible `MonitoringStatus.class` object. An external thread can also attempt to lock on the class object which could lead to a dead-lock situation. While it's unlikely it's still possible and this new implementation guarantees that won't happen due to the use of the private lock object (see _Effective Java_, Second Edition, _Item 70: Document thread safety_).

References:

Joshua Bloch. _Effective Java_, Second Edition. Addison-Wesley, 2008

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Bug Fixes:
 - Fix bug causing brief scan dropouts after starting a scan after a long period
   of inactivity (i.e. startup and background-foreground transitions) due to
   Android N scan limits (#489, Aaron Kromer)
+- Ensure thread safety for singleton creation of `BeaconManager`,
+  `DetectionTracker`, `MonitoringStatus`, and `Stats`. (#494, Aaron Kromer)
 
 
 ### 2.9.2 / 2016-11-22
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -108,7 +108,7 @@ import java.util.concurrent.CopyOnWriteArraySet;
 public class BeaconManager {
     private static final String TAG = "BeaconManager";
     private Context mContext;
-    protected static BeaconManager client = null;
+    protected static volatile BeaconManager client = null;
     private final ConcurrentMap<BeaconConsumer, ConsumerInfo> consumers = new ConcurrentHashMap<BeaconConsumer,ConsumerInfo>();
     private Messenger serviceMessenger = null;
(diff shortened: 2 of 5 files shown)


davidgyoung:
I've read conflicting literature about the usefulness of `volatile` in Java to address concurrency problems, but my guess is that this probably won't hurt.  I'd like to do a little testing with it to make sure I don't see it introducing any problems.

cupakromer:
I agree that `volatile` may not generally be overly useful. Here the use is meant to ensure the read in the first `if` condition shows the latest written value before entering the `synchronize` block (the source of that code is the _Effective Java_ Second Edition book).

Here's a bit more on `volatile`:

> When a field is declared `volatile`, the compiler and runtime are put on notice that this variable is shared and that operations on it should not be reordered with other memory operations. Volatile variables are not cached in registers or in caches where they are hidden from other processes, so a read of a volatile variable always returns the most recent write by any thread.
>
> ...
>
> Good uses of `volatile` variables include ensuring the visibility of their own state, that of the object they refer to, or indicating that an important life-cycle event (such as initialization or shutdown) has occurred.
>
> -- Brian Goetz. _Java Concurrency in Practice_. Addison-Wesley, 2006. page 38

davidgyoung:
Testing on Nexus 5X shows singletons are created as needed to detect beacons as expected.

PR #501, opened 2017-04-27
Minor updates

cupakromer:
No functionality change here. This simply cleans up some unused statements, spelling, and adds in nullability annotations. The nullability annotations help Android Studio better catch potential development time errors based on object presence assumptions.

This includes a simple comment addition about the thread safety of the classes. These classes are only used in one location currently. It's easier to manage thread safety as necessary in the encapsulating class. So this simply documents that this class isn't thread safe (i.e.  it's up to the client to ensure thread safety as necessary).

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/scanner/DistinctPacketDetector.java
+++ b/src/main/java/org/altbeacon/beacon/service/scanner/DistinctPacketDetector.java
@@ -22,7 +22,7 @@ public class DistinctPacketDetector {
     private static final int MAX_PACKETS_TO_TRACK = 1000;
 
     @NonNull
-    private Set<ByteBuffer> mDistinctPacketsDetected = new HashSet<>();
+    private final Set<ByteBuffer> mDistinctPacketsDetected = new HashSet<>();
 
     public void clearDetections() {
         mDistinctPacketsDetected.clear();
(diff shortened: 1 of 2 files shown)


src/main/java/org/altbeacon/beacon/service/scanner/DistinctPacketDetector.java
```diff
@@ -1,37 +1,40 @@
...
+import android.support.annotation.NonNull;
 
 import java.nio.ByteBuffer;
 import java.util.HashSet;
 import java.util.Set;
 
 /**
  * Created by dyoung on 4/8/17.
- *
+ * </p>
```
davidgyoung: I believe this is intended to be an open <p> tag

cupakromer: You are correct. I believe the older javadoc tools automatically added `<p>` tags for each paragraph but not the associated closing tag. Which is why I just added the closing tags. And then checking some other code auto formatted through older Android Studio it used `<p/>`. Now it looks like the latest AS now simply uses `<p>`.

I'll update. Thanks for catching this.

cupakromer:
Anything blocking this from being merged?

PR #507, opened 2017-05-10
Fix failure to stop scans

davidgyoung:
This fixes a failure to stop BLE scanning in two scenarios caused by #491:

1. When there is a nonzero between scan period.
2. When the unbind() is called on the BeaconManager to stop the BeaconService.

This was reported here:  http://stackoverflow.com/questions/43835705/how-to-properly-stop-scanning-beacon-using-android-beacon-library-altbeacon/43836157#43836157

Testing on Nexus 5X, showing logs after a call to
`BeaconReferenceApplication.this.unbindService(connection);`

Without change (note beacon packets are detected after service stops):

```
05-10 10:55:42.262 31353-31353/org.altbeacon.beaconreference E/BeaconService: onDestroy()
05-10 10:55:42.263 31353-31353/org.altbeacon.beaconreference D/BluetoothCrashResolver: stopped listening for BluetoothAdapter events
05-10 10:55:42.264 31353-31353/org.altbeacon.beaconreference D/BluetoothCrashResolver: Wrote 0 Bluetooth addresses
05-10 10:55:42.264 31353-31353/org.altbeacon.beaconreference I/BeaconService: onDestroy called.  stopping scanning
05-10 10:55:42.264 31353-31353/org.altbeacon.beaconreference D/CycledLeScanner: stop called
05-10 10:55:42.265 31353-31353/org.altbeacon.beaconreference D/CycledLeScanner: disabling scan
05-10 10:55:43.087 31353-31353/org.altbeacon.beaconreference D/CycledLeScannerForLollipop: got record
05-10 10:55:43.089 31353-32074/org.altbeacon.beaconreference D/BeaconParser: Ignoring pdu type 01
05-10 10:55:43.090 31353-32074/org.altbeacon.beaconreference D/BeaconParser: Processing pdu type FF: 02011a07ff4c0010020b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 with startIndex: 5, endIndex: 10
05-10 10:55:43.090 31353-32074/org.altbeacon.beaconreference D/BeaconParser: This is not a matching Beacon advertisement. (Was expecting 02 15.  The bytes I see are: 02011a07ff4c0010020b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

With change (note no beacon packets are detected after service stops):

```
05-10 10:52:43.784 26067-26067/org.altbeacon.beaconreference E/BeaconService: onDestroy()
[...]

PR #513, opened 2017-05-24
Scan documentation update and stop scan bug fix

cupakromer:
This is a follow up to #507 

## Bug Fix

- Fix where scanner thread is quit

  This fixes what I believe was a simply copy-paste error regarding how the scanner thread is quit. This uses the scanner thread handler, instead of the main thread handler, to schedule the `quit` operation. This way any queued scan jobs (such as a stop) complete before we quit the thread. This is what the comment states, but wasn't what the code implemented.

## Behavior Modification

Fix scan stop when bound but no monitored / ranged regions.

This adds a guard to ensure that we only attempt any further scans when the scan cycler is started. This helps ensure the cycler has terminated scanning for the case where all regions have been removed but the service is still bound. In this case only `stop` will be called (instead of both `stop` and `destroy`). When this happens there are two potential scheduled tasks still queued:

- `scanLeDevice(true)` from `deferScanIfNeeded`
- `scheduleScanCycleStop`

Which potential task is queued is based on if we were in an active scan period or a between scan period when `stop` was called. In the case of a between scan period the `deferScanIfNeeded` _will_ cause a faux scanning cycle to start up again when it ends (scanning won't actually start, but all flags and internal state appear as if it is running). It will run for one more scan cycle before it terminates in `finishScanCycle`. Similarly, if we are in an active scan period the scheduled job for finishing the scan cycle will continue for the duration of the scan period.

## Documentation / Annotations

This helps to try explaining how / when the various internal handlers should be used. This is important to help developers understand various threading flow orders, race conditions, and resource contention problems. As part of the documentation this adds `MainThread` and `WorkerThread` annotations to the deferred `Runnable` tasks' `run` methods.

This makes the current implicit assumption that the scan cycler is started/stopped on the main thread explicit. This helps quickly inform developers what the callback expectations are in regards to race conditions and state safety.

In order to verify this we need to trace the `start`/`stop`/`destroy` calls back to the service. The service's core methods `onCreate` and `onDestroy` are called on the main thread (see links below). This means the service creates it's `mMessenger` on the main thread, which creates the `IncomingHandler` on the main thread which binds to the main looper. However, the `IncomingHandler` class itself leaves the looper unspecified relying on looper for the thread which creates it.

As we did in #430, this modifies `IncomingHandler` to explicitly use the main looper as a future proofing mechanism. This way we can ensure this implicit expectation doesn't change or at least make it obvious when troubleshooting cases where someone expects it to have changed. Similarly, this adds the main thread annotation to it's `handleMessage` implementation.

Working from here we can confirm that the only places where the beacon monitoring/ranging is updated is from the main thread through the `IncomingHandler`. As the `IncomingHandler` is the main communication channel for the service we transfer the main thread expectation to the associated ranging/monitoring methods. _If_ someone decides to call these methods directly on the service these annotations help the IDEs check/document that such calls are expected from the main thread.

With all of that documented we can now confidently state that the scan cycler's `start`, `stop`, and `destroy` are also meant to be called from the main thread. As `start` and `stop` both call `scanLeDevice` we can start tracing any other threading expectations for it. We already know it's called from the main thread through deferred jobs. This leaves the `finishScanCycle` as the last call site.

`finishScanCycle` is only called from `scheduleScanCycleStop`. As this method name implies it's called through a deferred job on the main thread. It is also called the "first" time in `scanLeDevice`. Thus we've shown that `scanLeDevice`, `finishScanCycle`, and `scheduleScanCycleStop` are expected to run on the main thread.

## See Also

- https://developer.android.com/reference/android/os/Handler.html#Handler%28%29
- https://developer.android.com/training/multiple-threads/communicate-ui.html
- https://developer.android.com/reference/android/app/Service.html
- https://developer.android.com/guide/components/services.html
- #430
- https://developer.android.com/studio/write/annotations.html#thread-annotations

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -37,8 +37,10 @@ import android.os.Binder;
 import android.os.Build;
 import android.os.Handler;
 import android.os.IBinder;
+import android.os.Looper;
 import android.os.Message;
 import android.os.Messenger;
+import android.support.annotation.MainThread;
 import android.support.annotation.NonNull;
 
 import org.altbeacon.beacon.Beacon;
@@ -145,9 +147,16 @@ public class BeaconService extends Service {
         private final WeakReference<BeaconService> mService;
 
         IncomingHandler(BeaconService service) {
+            /*
+             * Explicitly state this uses the main thread. Without this we defer to where the
+             * service instance is initialized/created; which is usually the main thread anyways.
+             * But by being explicit we document our code design expectations for where things run.
+             */
+            super(Looper.getMainLooper());
             mService = new WeakReference<BeaconService>(service);
         }
 
+        @MainThread
         @Override
         public void handleMessage(Message msg) {
             BeaconService service = mService.get();
@@ -206,7 +215,7 @@ public class BeaconService extends Service {
      */
     final Messenger mMessenger = new Messenger(new IncomingHandler(this));
 
-
+    @MainThread
     @Override
     public void onCreate() {
         bluetoothCrashResolver = new BluetoothCrashResolver(this);
@@ -294,6 +303,7 @@ public class BeaconService extends Service {
         return false;
     }
 
+    @MainThread
     @Override
     public void onDestroy() {
         LogManager.e(TAG, "onDestroy()");
@@ -330,6 +340,7 @@ public class BeaconService extends Service {
     /**
      * methods for clients
      */
+    @MainThread
     public void startRangingBeaconsInRegion(Region region, Callback callback) {
         synchronized (rangedRegionState) {
             if (rangedRegionState.containsKey(region)) {
@@ -342,6 +353,7 @@ public class BeaconService extends Service {
         mCycledScanner.start();
     }
 
+    @MainThread
     public void stopRangingBeaconsInRegion(Region region) {
         int rangedRegionCount;
         synchronized (rangedRegionState) {
@@ -355,6 +367,7 @@ public class BeaconService extends Service {
         }
     }
 
+    @MainThread
     public void startMonitoringBeaconsInRegion(Region region, Callback callback) {
         LogManager.d(TAG, "startMonitoring called");
         monitoringStatus.addRegion(region, callback);
@@ -362,6 +375,7 @@ public class BeaconService extends Service {
         mCycledScanner.start();
     }
 
+    @MainThread
     public void stopMonitoringBeaconsInRegion(Region region) {
         LogManager.d(TAG, "stopMonitoring called");
         monitoringStatus.removeRegion(region);
@@ -371,6 +385,7 @@ public class BeaconService extends Service {
         }
     }
 
+    @MainThread
     public void setScanPeriods(long scanPeriod, long betweenScanPeriod, boolean backgroundFlag) {
         mCycledScanner.setScanPeriods(scanPeriod, betweenScanPeriod, backgroundFlag);
     }
--- a/src/main/java/org/altbeacon/beacon/service/BeaconService.java
+++ b/src/main/java/org/altbeacon/beacon/service/BeaconService.java
@@ -42,6 +42,7 @@ import android.os.Message;
 import android.os.Messenger;
 import android.support.annotation.MainThread;
 import android.support.annotation.NonNull;
+import android.support.annotation.WorkerThread;
 
 import org.altbeacon.beacon.Beacon;
 import org.altbeacon.beacon.BeaconManager;
@@ -391,6 +392,7 @@ public class BeaconService extends Service {
     }
 
     protected final CycledLeScanCallback mCycledLeScanCallback = new CycledLeScanCallback() {
+        @MainThread
         @TargetApi(Build.VERSION_CODES.HONEYCOMB)
         @Override
         public void onLeScan(BluetoothDevice device, int rssi, byte[] scanRecord) {
@@ -405,6 +407,7 @@ public class BeaconService extends Service {
             }
         }
 
+        @MainThread
         @Override
         public void onCycleEnd() {
             mDistinctPacketDetector.clearDetections();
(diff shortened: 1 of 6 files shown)


src/main/java/org/altbeacon/beacon/service/BeaconService.java
```diff
@@ -329,6 +339,7 @@ private PendingIntent getRestartIntent() {
     /**
      * methods for clients
      */
+    @MainThread
```
davidgyoung: Is this annotation intended to be used this way?  The [docs](https://developer.android.com/reference/android/support/annotation/MainThread.html) say it is supposed to indicate that the method should only be called on the main thread, and I don't see why that would be true in this case.  It seems to be intended for methods that manipulate the UI.  In this case, I don't think there is anything wrong from calling on a worker thread.  Perhaps a comment would be a better way to indicate that it is not necessarily run on a worker thread.

src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
```diff
@@ -193,14 +223,14 @@ public void run() {
...
-    @SuppressLint("NewApi")
+    @MainThread
     protected void scanLeDevice(final Boolean enable) {
         try {
             mScanCyclerStarted = true;
             if (getBluetoothAdapter() == null) {
                 LogManager.e(TAG, "No Bluetooth adapter.  beaconService cannot scan.");
             }
-            if (enable) {
+            if (enable && mScanningEnabled) {
```
davidgyoung: Not 100% sure why we would not want to execute this block if mScanningEnabled is false

src/main/java/org/altbeacon/beacon/service/BeaconService.java
cupakromer: The scanner classes are not thread safe. They are started / run from the beacon service which runs from the main thread. Additionally, in the scanner class `mHandler` is defined on the main looper. One of the deferred tasks which are queued on this handler call `scanLeDevice`. As things are right now, for all intents and purposes these methods need to run on the main thread.

While we could add this to the javadoc, the annotation documents this with the benefit of giving an extra hint to intellisense. Now if you write the following you'll get a lint warning/error depending on the IDE configuration:

```java
mScanHandler.post(new Runnable() {
    @WorkerThread
    @Override
    public void run() {
        scanLeDevice(true);
    }
}
```

src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScanner.java
cupakromer: Why would we want to allow a scan to begin after `stop` has already been called?

`mScanningEnabled` is the main "is this cycled scanner started or stopped" flag. It is only set in `start` and `stop`.

cupakromer:
I've attached some detailed method / state traces for the lollipop scanner as follows:

| Code | Stop on State | Log File |
|----|----|----|
| Before PR | Stop during between scan period | file: [pre_pr_between_stop.txt](https://github.com/AltBeacon/android-beacon-library/files/1027682/pre_pr_between_stop.txt) |
| Before PR | Stop during active scan period | file: [pre_pr_scan_stop.txt](https://github.com/AltBeacon/android-beacon-library/files/1027684/pre_pr_scan_stop.txt) |
| PR | Stop during between scan period | file: [pr_between_stop.txt](https://github.com/AltBeacon/android-beacon-library/files/1027685/pr_between_stop.txt) |
| PR | Stop during active scan period | file: [pr_scan_stop.txt](https://github.com/AltBeacon/android-beacon-library/files/1027683/pr_scan_stop.txt) |

davidgyoung:
What difference should I see in these traces?  For both the Before PR and PR files I see that scanning stops after STOP REQUESTED is annotated in the log, and there are no further scan starts.

cupakromer:
I've attached a new set of logs with more detailed annotations. I've also tried to better summarize the behavior in the chart here.

| Code | Cycle State | Log Behavior | Log Files |
|------|-------------|----------|----------|
| Pre  | Between     | Scanning stops but the between cycle continues, then scan cycle starts (but doesn't actually scan) though the flags state otherwise, then once the scan period is over things stops. | [between-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032758/between-beacons.txt) [between-no-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032755/between-no-beacons.txt)
| Pre  | Scan        | Scanning stops but the scan cycle continues until the end of the period. | [scan-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032754/scan-beacons.txt) [scan-no-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032757/scan-no-beacons.txt)
| PR   | Between     | Scanning stops and a single final `deferScanIfNeeded` is run. | [pr-between-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032756/pr-between-beacons.txt) [pr-between-no-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032753/pr-between-no-beacons.txt)
| PR   | Scan        | Scanning stops and a single final `scheduleScanCycleStop` is run. | [pr-scan-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032751/pr-scan-beacons.txt) [pr-scan-no-beacons.txt](https://github.com/AltBeacon/android-beacon-library/files/1032752/pr-scan-no-beacons.txt)


Essentially this PR stops extra unnecessary work  being performed by background tasks.

cupakromer:
@davidgyoung I've added more annotated versions of the log files. I think these help demonstrate the behavior a bit better.

davidgyoung:
Thanks for those tests, @cupakromer, those are helpful.  I'm going to do some more testing with these changes combined with #484 before merging, as that provides a good test case for starting and stopping CycledLeScanner.

cupakromer:
Is this good to merge to `master` or are there some changes I should make?

davidgyoung:
Sorry, I haven't been able to run my tests yet on Android O because I had trouble with the build on my test device.  I have the testing environment working now, so I will try to get the testing done tomorrow.

davidgyoung:
Did a bunch of tests yesterday on both Android 6.0 and Android O with this PR merged into the scheduled-job-scanning branch.  

I started and stopped the CycledLeScanner 12+ on each platform times shifting from background to foreground and scanning stopped appropriately and the thread was killed.  I also ran 32 10-second background scan cycles overnight on Android O and again saw scanning end appropriately.

Issue #512, opened 2017-05-20
Background region monitoring stability with 2.11-beta1


davidgyoung:
Library version 2.10 caused background monitoring to cause intermittent exit/entry events even when beacons are always around.

The problem was caused by a bug that unintentionally left low power scanning enabled at all times after a particular sequence.  This has been resolved in 2.11-beta in this change : 14c0508e8863a94e8a86ec4b3da3bb98a2c34de2 .  The problem did not exist in 2.9.x and earlier versions of the library.

Sequence to reproduce the problem:

1. Start the reference app
2. Turn on a beacon transmitting at 10 Hz or more so it is detected multiple times in a scan cycle (causes a region entry)
3. Turn off the beacon and wait for a region exit
4. Put the app to the background
5. Wait for a lower power scan to start in the background
6. Turn on the beacon so it is detected **during the low power scan** (causes a region entry)

After the above a region exit will happen periodically followed by an entry.  This is caused by the fact that OS-level low power scanning has its own duty cycle that will delay results, and because the library does not take this into account, it will cause region exits inappropriately if results don't come quite quickly enough.

You  will know you have entered this state with 2.10 after seeing the debug line "We've been detecting for a bit.  Stopping Android L background scanning".

For this test, the following branch of the reference app (modified to use 2.10) was installed and run on  a  Nexus 9 on Android 7.0 with the app always running in the background.  https://github.com/AltBeacon/android-beacon-library-reference/tree/multi-process-support

```
adb logcat | grep -e "BeaconReferenceApp:" -e "beaconService version" -e "Stopping Android L background scanning"

05-20 19:31:34.676 28386 28386 D BeaconReferenceApp: this is the main process
05-20 19:31:34.676 28386 28386 D BeaconReferenceApp: setting up background monitoring for beacons and power saving
05-20 19:31:34.698 28386 28386 I BeaconService: beaconService version 2.10 is starting up
05-20 19:31:52.575 28386 28717 D BeaconReferenceApp: did enter region.
05-20 19:31:52.575 28386 28717 D BeaconReferenceApp: auto launching MainActivity
05-20 19:32:46.969 28386 29510 D BeaconReferenceApp: did exit region.
05-20 19:33:04.563 28386 29795 D BeaconReferenceApp: did enter region.
05-20 19:33:04.563 28386 29795 D BeaconReferenceApp: Sending notification.
[...]

PR #515, opened 2017-05-25
Log more specific error message on scan failure

cupakromer:
On occasion I noticed a "Scan Failed" in the logs. This improves the logged error message to report the associated reason based on the error code. If it's an error we don't know about it includes the error code in the message for further research.


davidgyoung:
I've seen this too.  Have you happened to capture any results after making this change?

cupakromer:
I should have more info soon. I think it may be related to an issue I'm seeing where regions are flapping in the background. I'm currently investigating.

cupakromer:
@davidgyoung so the scan failure I had seen was:

```
Scan failed: a BLE scan with the same settings is already started by the app
```

This was related to the flapping I was seeing which turns out to be the same as issue #512.

PR #516, opened 2017-05-26
Misc minor fixes

cupakromer:
This is mostly just a nullability annotation update with a few minor code changes. However, there were two sets of significant changes which actually fixed a few potential bugs:

- Fix possible `NullPointerException` with `BackgroundPowerSaver` on devices prior to Android 4.3 Jelly Bean MR 2 (API 18)

  This eliminates the short circuit in the guard clause which would prevent a `BeaconManager` from being set. The basic creation of a `BeaconManager` instance doesn't require Android 18; running that instance will and is turned into a no-op at that point. While here nullability annotations were included and the `BeaconManager` instance was made `final` (this is how the NPE was caught).

- Fix rare edge case causing `NoSuchElementException` when using the legacy `BeaconManager#getMonitoringNotifier` and `BeaconManager#getRangingNotifier` where the notifier sets were modified external to `BeaconManager` by another thread

  The set of range and monitor notifiers is stored in a thread safe `CopyOnWriteArraySet` instance. Adding an additional layer of synchronization around already atomic operations is unnecessary. The only case where we performed a non-atomic action is in the legacy `getRangingNotifier` and `getMonitoringNotifier` implementations.

  In these legacy cases, synchronization is necessary to ensure that the empty set check isn't invalid by the following iterator access (which would throw a `NoSuchElementException`). The iterator creation is necessary because unlike lists sets don't allow elements to be requested by index. But, if we drop the synchronization for all of the other use cases, while leaving it here, doesn't provide us this guarantee.

  However, we actually don't even have that guarantee now. This is because the set is publicly exposed through the getters `getRangingNotifiers` and `getMonitoringNotifiers`. Any client can request and manipulate the sets external to the manager potentially causing a confusing `NoSuchElementException` exception.

  One of the reasons for using `CopyOnWriteArraySet` is that their iterators are thread safe allowing unsynchronized callback iteration by the beacon service. This drops the explicit synchronization and corrects the legacy getter implementations by only using the iterator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Bug Fixes:
 
 - Fix failure to stop scanning when unbinding from service or when the between scan period
   is nonzero. (#507, David G. Young)
+- Fix possible `NullPointerException` with `BackgroundPowerSaver` on devices
+  prior to Android 4.3 Jelly Bean MR 2 (API 18) (#516, Aaron Kromer)
 
 ### 2.10 / 2017-04-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Bug Fixes:
   is nonzero. (#507, David G. Young)
 - Fix possible `NullPointerException` with `BackgroundPowerSaver` on devices
   prior to Android 4.3 Jelly Bean MR 2 (API 18) (#516, Aaron Kromer)
+- Fix rare edge case causing `NoSuchElementException` when using the legacy
+  `BeaconManager#getMonitoringNotifier` and `BeaconManager#getRangingNotifier`
+  where the notifier sets were modified external to `BeaconManager` by another
+  thread (#516, Aaron Kromer)
 
 ### 2.10 / 2017-04-21
 
--- a/build.gradle
+++ b/build.gradle
@@ -56,8 +56,12 @@ android {
         targetSdkVersion 23
         versionCode 1
         versionName version
-        testInstrumentationRunner "com.google.android.apps.common.testing.testrunner.GoogleInstrumentationTestRunner"
         consumerProguardFiles 'proguard-rules.pro'
+        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        testOptions {
+            // The test runner needs Espresso 2.2.2 which requires at least 8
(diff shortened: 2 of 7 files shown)


davidgyoung:
> This is because the set is publicly exposed through the getters getRangingNotifiers and getMonitoringNotifiers. Any client can request and manipulate the sets external to the manager potentially causing a confusing NoSuchElementException exception.

We could solve this problem by making a copy of the set before returning it, e.g.:

```
public Collection<Region> getRangedRegions() {
    return new HashSet<>(rangeNotifiers)
}
```

cupakromer:
> We could solve this problem by making a copy of the set before returning it

Sure happy to make that change. I wasn't sure if that would be considered a "breaking behavior" change as they returned the raw sets when implemented in #392. A few follow up thoughts:

- Providing a copy wouldn't completely fix the underlying thread concern so the new `Iterator` based implementation for the single getter is still required
- Returning a copy would further ensure, but not completely as these objects are listed as `protected` so subclass can modify them at will, that any updates to the set used by the manager are not missed 👍 
- These getters are also used internally by `BeaconIntentProcessor`; do you have any concerns about the additional overhead of creating an additional copy of the list given they are internally thread safe `CopyOnWriteArraySet` instances

  If this is a concern, given that the underlying `rangeNotifiers` and `monitorNotifiers` objects are `protected` we could access them directly from `BeaconIntentProcessor`. Or per the 2nd bullet point if them being `protected` is not the long term goal for data safety we could add a package private alternative getter which provides access - and/or even make the fields package private as well.

davidgyoung:
I don't consider making a copy of the set is a significant breaking API change.  It's hard to imagine cases where this would affect users.  Besides, the getter has not been public in non-beta  release yet.

The `BeaconIntentProcessor` will fire this getter once per second when ranging in the foreground under default settings, so a copy of the set will be made fairly often.  I'm not super worried about this, since the number of notifiers is typically low and this copy will not use a lot of processing power.  Probably still better than accessing the field directly.

cupakromer:
> I don't consider making a copy of the set is a significant breaking API change. It's hard to imagine cases where this would affect users. Besides, the getter has not been public in non-beta release yet.

👍  Sounds good. @davidgyoung made the change and updated the javadocs for those getters to better explain usage.

PR #523, opened 2017-06-03
Fix `RunningAverageRssiFilter.setSampleExpirationMilliseconds`

davidgyoung:
The ability to configure this setting was broken as reported in https://github.com/AltBeacon/android-beacon-library-reference/issues/30.  This change makes the API work as intended so the documentation is correct.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
+++ b/src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
@@ -17,7 +17,7 @@ public class RunningAverageRssiFilter implements RssiFilter {
 
     private static final String TAG = "RunningAverageRssiFilter";
     public static final long DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS = 20000; /* 20 seconds */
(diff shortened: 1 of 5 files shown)


src/main/java/org/altbeacon/beacon/service/RunningAverageRssiFilter.java
```diff
@@ -17,7 +17,7 @@
 
     private static final String TAG = "RunningAverageRssiFilter";
     public static final long DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS = 20000; /* 20 seconds */
-    private static long sampleExpirationMilliseconds = DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS;
+    protected static long sampleExpirationMilliseconds = DEFAULT_SAMPLE_EXPIRATION_MILLISECONDS;
```
cupakromer: It looks like this may only be needed to have higher access for testing. Instead of exposing this as `protected`, which makes it exposed in public documentation, could we keep it `private` and use a getter with package-private access instead?

```java
    @RestrictTo(TESTS)
    static long getSampleExpirationMilliseconds() {
        return sampleExpirationMilliseconds;
    }
```

PR #536, opened 2017-07-05
Fixed getRunningAverageRssi Beacon class method

peterdeka:


davidgyoung:
Thanks, @peterdeka.  I added a commit to your branch to keep the old method but deprecate it.  This will keep a library upgrade from breaking anybody's code if they somehow happened to use the old bad method definition.

Issue #561, opened 2017-08-14
NPE when initialising while bluetooth is turned off under Android O preview


strooooke:
Instantiating and configuring a `BeaconManager` while bluetooth is turned off crashes in https://github.com/AltBeacon/android-beacon-library/blob/53ccb8c752f1ac7e7fdfb81ac3ad1f5a5f6e1ff4/src/main/java/org/altbeacon/beacon/service/ScanHelper.java#L185  

When bluetooth is off, `bluetoothAdapter` is not null, but `bluetoothAdapter.getBluetoothLeScanner()` is. Alternatively, check for `bluetoothAdapter.isEnabled()`.

library version is 2.12; observable on both Nexus 5X and Pixel with Android O beta, e.g. by turning bluetooth off and starting the demo app.

PR #562, opened 2017-08-14
Fix NPE in ScanJob on Android O while bluetooth is turned off

strooooke:
Fixes https://github.com/AltBeacon/android-beacon-library/issues/561


davidgyoung:
This looks good, @strooooke.  Sorry I didn't catch this case myself.  I'll run some tests on my Android 8 device with bluetooth on and off before merging, then I'll roll a new release before Android 8 goes into the wild.

Issue #593, opened 2017-09-15
java.lang.NullPointerException at ProcessUtils.java:23


TonyTangAndroid:
### Expected behavior
No Exception
### Actual behavior
Null Pointer Expcetion
### Steps to reproduce this behavior
Not known
### Mobile device model and OS version
LG G4 Android 6.0
### Android Beacon Library version
2.12.2
Stack Trace:

> Fatal Exception: java.lang.RuntimeException: Unable to start receiver org.altbeacon.beacon.startup.StartupBroadcastReceiver: java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2750)
       at android.app.ActivityThread.access$1800(ActivityThread.java:157)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1433)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:148)
       at android.app.ActivityThread.main(ActivityThread.java:5525)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)
Caused by java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
       at org.altbeacon.beacon.utils.ProcessUtils.getProcessName(ProcessUtils.java:23)
       at org.altbeacon.beacon.BeaconManager.checkIfMainProcess(BeaconManager.java:361)
       at org.altbeacon.beacon.BeaconManager.<init>(BeaconManager.java:319)
       at org.altbeacon.beacon.BeaconManager.getInstanceForApplication(BeaconManager.java:310)
       at org.altbeacon.beacon.startup.StartupBroadcastReceiver.onReceive(StartupBroadcastReceiver.java:32)
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2743)
       at android.app.ActivityThread.access$1800(ActivityThread.java:157)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1433)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:148)
[...]

PR #598, opened 2017-09-21
Protect against NPE on process util process check

davidgyoung:
Per report in #593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Development
 
 Bug Fixes:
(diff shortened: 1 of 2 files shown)


ramanpreetSinghKhinda:
In which AltBeacon library version this issue is fixed? Is this `2.12.4`?

davidgyoung:
Per the changelog [here](https://github.com/AltBeacon/android-beacon-library/blob/master/CHANGELOG.md), this was in the 2.12.3 release.

PR #600, opened 2017-09-22
Fix isAnyConsumerBound method

carlonzo:
We should check that the Map is not empty.

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -495,7 +495,7 @@ public class BeaconManager {
      */
     public boolean isAnyConsumerBound() {
         synchronized(consumers) {
-            return consumers.isEmpty() && (serviceMessenger != null);
+            return !consumers.isEmpty() && (serviceMessenger != null);
         }
     }
 


hashbrown:
I would like to suggest that the problem is a bit broader than this, and would love to see it fixed in this PR.

For backward compatibility reasons (I believe) when using the the new Android 8 impl (2.12.x), when you attempt to bind to the BeaconService via BeaconManager.bind() it simply invokes the onBeaconConnected() callback, and gives the appearance to your app of completing the service binding.
However, if your existing code expects that you actually were bound to the service, and uses the isAnyConsumerBound() or isBound(BeaconConsumer) methods, they will always return false, since there is no service connection established.

I think this represents a semantically breaking api change of BeaconConsumer.

My proposal is that if scheduledScanJobsIsEnabled == true (using new Android 8 strategy) then the api contract is kept consistent, meaning that any isBound() calls, only consider the presence of a consumer and ignore the service connection condition.

davidgyoung:
@hashbrown, I agree with your proposal.  But based on @carlonzo's PR, that method was completely broken before (the logic was reversed), so the contract was already broken.  That said, the right thing to do is fix it as you describe.

So I believe the line should now look like this:

```return scheduledScanJobsIsEnabled == true || (!consumers.isEmpty() && (serviceMessenger != null));```

**Comment by hashbrown** [comment]
Actually I think it would be more accurate to say
```return !consumers.isEmpty() && (scheduledScanJobsIsEnabled || serviceMessenger != null)```

And then
```
public boolean isBound(@NonNull BeaconConsumer consumer) {
        synchronized(consumers) {
            // Annotation doesn't guarantee we get a non-null, but raising an NPE here is excessive
            //noinspection ConstantConditions
            return consumer != null && consumers.get(consumer) != null && (scheduledScanJobsIsEnabled || serviceMessenger != null);
        }
    }
```

davidgyoung:
Sounds good to me, @hashbrown.  Do you want to prepare a new PR based on what is discussed above?

hashbrown:
Certainly.

geeksunny:
Any updates on this? I'm running into the same issue as #613 and need to figure out a workaround as our app won't scan under Android 8.0 right now.

davidgyoung:
Replacement PR with changes described above is #655

Issue #612, opened 2017-10-25
Identifier parsing creates to many new objects


ekryvoshey:
Hello!

I have a problem: my BeaconLayout is like this "m:0-1=7502,p:3-3,i:0-1,i:4-14,i:15-24". Identifiers summary length is pretty long, and it causing Indetifier object to create too many bew Object during toHexString() execution. And that causing to many garbage collection in the app, UI thread is constantly interrupting with GC so I have UI lag every time I'm receiving beacon colection.

https://drive.google.com/open?id=0B_546mjMnsJ9d2QyazJyc1FraU0

As far as I remember, data fields could be an alternative, but aren't they limited to 4 bytes? Perhaps you can suggest different solution to this?

davidgyoung:
Yes, data fields are limited to four bytes.

It sounds like the proper answer is to simply optimize the toHexString() method of the Identifier class so toString() isn't so expensive for identifiers that aren't two bytes or 16 bytes.

At first glance simply replacing `String.format("%02x", item)` with code that doesn't use the heavy-weight String.format method might improve things quite a bit.  Here's one such proposed implementation: https://stackoverflow.com/a/32123766/1461050

@ekryvoshey, do you want to try benchmarking a replacement method implementation to see if you can come up with something faster?  If so, I'll put it into a test release for you to try.

ekryvoshey:
Yes, I would like to try it.

ekryvoshey:
Yes, I would like to try and figure out the better solution. If you can assist, it would be great!

PR #615, opened 2017-10-27
Fix performance problems caused by Identifier#toHexString

davidgyoung:
This addresses performance problems described in #612 where usage of beacon layouts with 3-15 byte identifiers results in slow calls to Identifier#toHexString, as this is what is used to convert identifiers of this length range to Strings.

Unfortunately, it is very difficult to remove the use of `toHexString()` in normal operation of the library, because it is used for both serialization of `Beacon` objects when moving them in the library, and for the implementation of `hashCode()` in `Beacon`, which is heavily used for internal beacon tracking.

Two changes:

1. Wrap two debug level logging calls to `beacon.toString()` in checks to see if verbose logging is enabled, so in normal cases, there will be no call ` toString()`  and hence no call to `toHexString()` on `Identifier`.

2. Optimize the implementation of `toHexString()` so it does not instantiate objects in a loop and instead uses primitive types for all conversion parts except the final construction of a single String object.

The first change reduces the number of calls to `toHexString()` by 51%, with all the remaining calls being used for serialization and hashCode calculations.

```
Before disabling verbose logging:
toHexString calls total: 366, serialize: 12, hash: 168

After disabling verbose logging:
toHexString calls total: 442, serialize: 34, hash: 408
```

The second change improves the performance of the method by 90% from 240 microseconds per call to 23 microseconds per call on a Nexus 5X.

```
Original implementation:
10-27 14:31:56.862 15936 15974 I Identifier: toHexString callCount: 558 Nanos per call: 241237

New implementation:
10-27 15:06:01.371 25842 26260 I Identifier: toHexString callCount: 722 Nanos per call: 22948 
```

Between these two changes, processing time in toHexString() is reduced by 95%, and the number of objects allocated is sharply reduced.


davidgyoung:
A test build with these changes, along with instructions on how to configure it with your project is here: 
 https://github.com/AltBeacon/android-beacon-library/releases/edit/optimize-identifier-hex-conversion

PR #617, opened 2017-11-05
Fix NPE on process background detection due to race condition

### LINKED ISSUE #616: NPE Background Crash in ScanHelper#processBeaconFromScan
### Actual behavior
* App crashed in background due to NPE
* The phone showed a popup saying the app which uses altbeacon has crashed, but this alert dismissed itself after less than a second without interaction on my part.

I don't know reproducible steps. At the time this happened, the following things were happening:
* The phone was connected to my car's bluetooth which uses HFP, A2DP, and AVRCP protocols
* I was playing music with Spotify over the Bluetooth
* Two Tile beacons would have been in range
* I hadn't run the app that uses AltBeacon for several days

Pixel 2 (walleye), Android 8.0

2.12.2

davidgyoung:
Fixes #616

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+### Development
+
+Bug Fixes:
+
+ - Fix NullPointerException in ScanHelper#processBeaconFromScan (#617, David G. Young)
+
 ### 2.12.3 / 2017-10-14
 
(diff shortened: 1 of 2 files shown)


davidgyoung:
closed in favor of #626

PR #621, opened 2017-11-19
Fix missing runningAverageRssi values

davidgyoung:
This change adds missing serialization of the `runningAverageRssi` field on `Beacon`.  The fact that this was missing was inadvertent.   As a result of this being missing, the running average value was not used in distance calculations, instead falling back to using the most recent single measurement.  This greatly impacts the noise on distance estimates.  This defect was likely  introduced by accident in another change.

Additionally, this change adds two packet collection metrics to the `Beacon` class:

* `packetCount` (tells how many packets have been detected for beacon since ranging started)
* `rssiMeasurementCount` (tells how many rssi samples went into the runningAverageRssi measurement, or 0 if indeterminable for the RssiFilter implementation)


cupakromer:
Just noticed this after I made a similar change in #630. Per 1f02531c4 what I found was that the running average is available when using the scan job logic but not when using the service intent. Happy to adjust things to make these to PRs work with each other.

davidgyoung:
Yes, @cupakromer, if you can see if you can make the two work together, I'd appreciate it.  I'll merge this now to make it easier, and then you can pul the changes in to your branch.

PR #631, opened 2017-12-13
Minor refactor and annotation updates

cupakromer:
This adds annotations to `ExtraDataBeaconTracker` and it's use within `ScanHelper`. This came about after reviewing #617 and #626. While trying to understand how `ExtraDataBeaconTracker` was processing the beacons I ended up re-arranging some of the code to flatten the conditionals and reduce the number of branches I had to track mentally resulting in a minor refactor. This also lead to noticing that the setting of the tracked beacon extra data wasn't tested, so the tests were updated to check this.


davidgyoung:
Annotations and tests look good.  I just merged #616, so we don't lose any changes.  Perhaps you can merge the latest from master so everything is in here.

PR #640, opened 2018-01-18
Fix CHANGELOG.md

ekatek:
Change 2.13.4 to 2.12.4, which is the version that it actually appears to be.

Hey y'all, so I just spent a while confused about adding 2.13.4 to my project and it didn't seem to exist. Since the previous project version in the CHANGELOG is 2.12.4 are you sure that you didn't mean 2.12.4? :)


davidgyoung:
Good catch!  Clearly a typo.

PR #645, opened 2018-02-18
Configurable scan job ids

### LINKED ISSUE #582: Used jobIds should be documented and configurable
See https://commonsware.com/blog/2017/06/07/jobscheduler-job-ids-libraries.html for a much better written explanation of why it matters.

It's up to the client app to make sure that their chosen jobIds will then be unique inside their uid, and stay fixed over process restarts, device reboots, updates etc... a use case for statics if I've ever seen one. But they need to be configurable to avoid clashes.

davidgyoung:
The Job Scheduler ScanJobs used in Android 8 has a hard-coded job id of 1 and 2.  Because this job id must be unique per application, this can cause collisions with other app job ids.  This changes the default job ids to be a high integer number, and declares them in AndroidManifest.xml:

```
        <service android:name=".service.ScanJob"
                android:permission="android.permission.BIND_JOB_SERVICE" >
            <meta-data android:name="immmediateScanJobId" android:value="208352939" />
            <meta-data android:name="periodicScanJobId" android:value="208352940" />
        </service>
```

These may be overridden in the app manifest as needed like this: 

```
        <service android:name="org.altbeacon.beacon.service.ScanJob">
            <meta-data android:name="immediateScanJobId" android:value="1000" tools:replace="android:value"/>
            <meta-data android:name="periodicScanJobId" android:value="1001" tools:replace="android:value"/>
        </service>
```

The job ids may also be changed programmatically with: 

```
        ScanJob.setOverridePeriodicScanJobId(100);
        ScanJob.setOverrideImmediateScanJobId(101);
```

However, for third party apps using the library the programmatic syntax should be avoided as it accesses a private API and is not guaranteed to be supported in future releases.

This resolves #582

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ### Development
 
-Bug Fixes:
+Enhancements:
+ - Allow configuring job ids (#645, David G. Young)
+
+ Bug Fixes:
  - Allow scans with screen off on Android 8.1 (#637, David G. Young)
 
 ### 2.13.4 / 2017-12-16
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -30,8 +30,8 @@
 
         <service android:name=".service.ScanJob"
                 android:permission="android.permission.BIND_JOB_SERVICE" >
-            <meta-data android:name="immmediateScanJobId" android:value="208352937" />
-            <meta-data android:name="periodicScanJobId" android:value="208352938" />
+            <meta-data android:name="immmediateScanJobId" android:value="208352939" />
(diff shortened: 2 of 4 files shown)


dietert:
Hi David

I suppose that "immmediateScanJobId" with 3 'm'-s is a typo in the documentation?

Regards
Dieter

davidgyoung:
Yes, this is a typo in the documentation which I just fixed [here](https://github.com/AltBeacon/android-beacon-library/commit/4184356d407164a931da37f36db46fcc682c802a).  This same typo actually was in the AndroidManifest.xml for the library when it was released in version 2.13, but it was quickly updated in minor release 2.13.1 in #653

Issue #641, opened 2018-01-26
Android O ranging problem


FongMi:
### Expected behavior
Android O (8.0) can search beacon in my regions

### Actual behavior
The log says
> BeaconManager started up on pid 7266 named 'com.direk.android.newest' for application package 'com.direk.android.newest'.  isMainProcess=true
> This consumer is not bound.  Binding now: com.direk.android.clock.ClockBeaconActivity@fbefc
> Not starting beacon scanning service. Using scheduled jobs
> consumer count is now: 1

### Steps to reproduce this behavior
Same as Ranging Example Code
http://altbeacon.github.io/android-beacon-library/samples.html

### Mobile device model and OS version
HTC_M10
Android 8.0

### Android Beacon Library version
android-beacon-library-2.12.4

Is there have any example for Android O ?

davidgyoung:
@FongMi, the log lines shown look normal.   Can you please describe the problem in greater detail?

Do you ever get a ranging callback on Android 8?
Does your code work on Android 4.3-7.x devices?
How long to do you wait for a ranging callback on Android 8?
Does it make a difference in the foreground vs. the background?
Can you detect your beacon using an off the shelf beacon detector?
Have you added the proper beacon layout for your beacon (if it is iBeacon or Eddystone)?

FongMi:
Do you ever get a ranging callback on Android 8?
->no

Does your code work on Android 4.3-7.x devices?
->yes

How long to do you wait for a ranging callback on Android 8?
->I have not get the callback

Does it make a difference in the foreground vs. the background?
->I only use foreground scan

Can you detect your beacon using an off the shelf beacon detector?
->Sure

Have you added the proper beacon layout for your beacon (if it is iBeacon or Eddystone)?
->Yep, I add a iBeacon layout by myself.
->m:2-3=0215,i:4-19,i:20-21,i:22-23,p:24-24

davidgyoung:
Thanks for the answers, @FongMi .  This certainly is not expected behavior and is not consistent with tests I have done with the reference app on Android 8.  Would you mind running the reference app on your device and see if you see the same?

https://github.com/AltBeacon/android-beacon-library-reference

I wonder if there is some difference with the way your app is using the library that makes it incompatible with Android 8.

FongMi:
ok, I will try again, thanks.

MahmoudElFerekh:
[...]

Issue #613, opened 2017-10-26
beaconManager.isBound() does not work in Android O


mgsteinkamp:
### Expected behavior
After onBeaconServiceConnect(), beaconManager.isBound() will evaluate as true
### Actual behavior
After onBeaconServiceConnect(), beaconManager.isBound() evaluates as false
### Steps to reproduce this behavior
```
beaconManager.bind(BeaconConsumer b);
...
onBeaconServiceConnect();
...
Log.i(TAG, "Bound: " + beaconManager.isBound());
```

### Mobile device model and OS version
Nexus 5X

### Android Beacon Library version
2.12.3

###Notes:
Worked totally fine on test devices from Android 5-7.1
Possibly related: RegionBootstrap doesn't seem to function on the first set any more either, but sorry, I don't have time to file a separate report.

davidgyoung:
Thanks for this report.  A proposed solution is described in #600

davidgyoung:
Fixed in #655

PR #655, opened 2018-03-06
Fix flawed logic on BeaconManager#isBound

davidgyoung:
This addresses the issue reported in #613 where isBound and isAnyConsumerBound don't work with schedule jobs (Android 8+ default).  This also reportedly caused trouble in #641.

This is a proposed replacement for pull request #600

@hashbrown  and @carlonzo, both of you proposed fixes for this problem, so any comments are welcome.


geeksunny:
@davidgyoung I tested this fix in my app and it solved the issue preventing scanning from working on Android 8.0 without any code changes. I'll sign off on this logic. 👍

davidgyoung:
Thanks for the check @geeksunny

PR #663, opened 2018-03-25
Allow beacon service to run as foreground service

davidgyoung:
Adds the ability to configure the beacon scanning service as a foreground service.  This is especially useful on Android 8+ for applications that need scans to run more often than every 15 minutes and are OK with an icon being displayed to the user while the app is doing this.

Here's an example of setting this up:

        Notification.Builder builder = new Notification.Builder(this);
        builder.setSmallIcon(R.mipmap.my_icon);
        builder.setContentTitle("Scanning for Beacons");
        Intent intent = new Intent(this, MyMainActivity.class);
        PendingIntent pendingIntent = PendingIntent.getActivity(
                this, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT
        );
        builder.setContentIntent(pendingIntent);
        beaconManager.enableForegroundServiceScanning(builder.build(), 456);

At a later time you can call:

        beaconManager.disableForegroundServiceScanning();

But both calls to enable and disable must be made when the service is not bound, so you have to stop ranging/monitoring first then call `beaconManager.unbind(...);` before disabling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+### 2.14 / 2018-05-17
+
+Enhancements:
+
+ - Add warnings about disabling scheduled scan jobs on Android 8+ (#674, David G. Young)
+ - Add warning about setting short scan periods on Android 8 (#677, David G. Young)
+ - BeaconTransmitter advertisements may be configured as connectable (#683, Michael Harper)
+
+Bug Fixes:
+ - Fix crashes of BluetoothMedic caused by Bluetooth being turned off (#675, David G. Young) 
+ - Fix flawed in BeaconManager.isBound method on Android 8 (#655, David G. Young)
+
 ### 2.13.1 / 2018-03-05
 
 [Full Changelog](https://github.com/AltBeacon/android-beacon-library/compare/2.13.1...2.13)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### Development
+
+Enhancements:
+
+ - Optional foreground beacon scanning service for faster background detections on Android 8+
+   (#663, David G. Young)
+
 ### 2.14 / 2018-05-17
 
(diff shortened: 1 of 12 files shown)


twanschik:
Hi David,
I would prefer to use the default notification builder inside my code to specify the notification. This will allow me to specify the notification dynamically i.e. to change the title or the icon at runtime. This is much more flexible than referencing these attributes statically in the manifest.

What do you think?

HughIngram:
This would be helpful in my use case. 

I agree with @twanschik - it would be preferable to specify (and modify) the notification at runtime; my application already uses a foreground notification, so I would like to be able to provide BeaconService with the same Notification ID, so that my service and BeaconService will share the same notification.

davidgyoung:
@twanschik ad @HughIngram, I have reworked this to allow programmatic configuration of the foreground service notification rather than declarative configuration in the manifest.   I am a little reluctant to keep adding complexity to the BeaconManager, but I think that is the proper way to provide this important feature under programmatic control.

snydersaurus:
When I run this on my Android 6.0 device, I see the notification tray icon; however it is not present when running an Android 8.0 device. From the description, I'd expect to see the icon. Am I missing something?

davidgyoung:
@snydersaurus, what device do you see this on?  I noticed something similar on a Samsung Galaxy S9 with 8.0, but I figured it was a Samsung customization not an Android 8 thing.

josemssilva:
Any clue as to when this will be merged to master? Thanks.

davidgyoung:
@josemssilva, I'd like to mere this later this week, but I'd really like to get some confirmation that other folks have had success testing with the branch. 

Can anybody confirm they have had this branch work successfully?  I'd like to know it's not just me who has had success.

josemssilva:
I haven't done a lot of tests, but regarding the ones I've made everything went as expected.

Issue #673, opened 2018-04-24
"Not allowed to start service Intent" crashes when app is in background


scm573:
Hi guys,

My app is suffering crashes when in background, on Android 8.0. (I've not tried on 8.1 though)

### Expected behavior
App can scan beacons and receive events when in background

### Actual behavior
App crashes when the library executes its callback.

### Steps to reproduce this behavior
1. Run the app on a device approximated to an active beacon.
2. Press home button and wait for its scanning.
3. Crash.

### Mobile device model and OS version
Nexus 5X / Sony Xperia Z1 Compact
Android 8.0.0

### Android Beacon Library version
2.13.1

Here's the stack trace:

> 04-24 11:30:10.010 9868-9868/com.company.product E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.company.product, PID: 9868
    java.lang.IllegalStateException: Not allowed to start service Intent { cmp=com.company.product/org.altbeacon.beacon.BeaconIntentProcessor (has extras) }: app is in background uid UidRecord{27a0755 u0a265 LAST bg:+1m23s166ms idle procs:1 seq(0,0,0)}
        at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1505)
        at android.app.ContextImpl.startService(ContextImpl.java:1461)
        at android.content.ContextWrapper.startService(ContextWrapper.java:644)
        at org.altbeacon.beacon.service.Callback.call(Callback.java:60)
        at org.altbeacon.beacon.service.BeaconService.processExpiredMonitors(BeaconService.java:368)
        at org.altbeacon.beacon.service.BeaconService.access$200(BeaconService.java:69)
        at org.altbeacon.beacon.service.BeaconService$1.onCycleEnd(BeaconService.java:314)
        at org.altbeacon.beacon.service.scanner.CycledLeScanner.finishScanCycle(CycledLeScanner.java:242)
        at org.altbeacon.beacon.service.scanner.CycledLeScanner.scheduleScanCycleStop(CycledLeScanner.java:234)
        at org.altbeacon.beacon.service.scanner.CycledLeScanner$1.run(CycledLeScanner.java:230)
        at android.os.Handler.handleCallback(Handler.java:789)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:164)
        at android.app.ActivityThread.main(ActivityThread.java:6541)
        at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #674, opened 2018-04-25
add warning about disabling scan jobs on Android 8+

davidgyoung:
This is a documentation and LogCat warning change to address the concern in  #673 that calling `setEnableScheduledScanJobs(false)` can break functionality on Android 8+

### DIFF
--- a/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -548,6 +548,10 @@ public class BeaconManager {
      * otherwise beacon scans may be run only once every 15 minutes in the background, and no low
      * power scans may be performed between scanning cycles.
      *
+     * Setting this value to false will disable ScanJobs when the app is run on Android 8+, which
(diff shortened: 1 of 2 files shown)


src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -569,8 +573,13 @@ public void setEnableScheduledScanJobs(boolean enabled) {
                     " availble prior to Android 5.0");
             return;
         }
+        if (enabled && android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
mannodermaus: Quick question: If this is concerned about logging a warning when _disabling_ scan jobs, shouldn't this condition be inverted, and check for `!enabled` instead?

davidgyoung: Good catch, thanks!  It should indeed be !enabled.  Care to open an issue (or for bonus points a PR)?

mannodermaus: Sure! Raised as #700

Issue #141, opened 2015-03-27
How to set the Advertising Flags when simulating a Beacon


michael-customlbs:
I got my test application to send out iBeacon packets. Most Apps recognize them, but some don't.

I saw that the packets are missing some flags (3 bytes in the beginning to be exact).
https://developer.mbed.org/blog/entry/BLE-Beacons-URIBeacon-AltBeacons-iBeacon/

Most iBeacons I know of send out 0x020106 or 0x02011A (see example below) as the first 3 bytes.

	02 # Number of bytes that follow in first AD structure
	01 # Flags AD type
	1A # Flags value 0x1A = 000011010  
	   bit 0 (OFF) LE Limited Discoverable Mode
	   bit 1 (ON) LE General Discoverable Mode
	   bit 2 (OFF) BR/EDR Not Supported
	   bit 3 (ON) Simultaneous LE and BR/EDR to Same Device Capable (controller)
	   bit 4 (ON) Simultaneous LE and BR/EDR to Same Device Capable (Host)
	1A # Number of bytes that follow in second (and last) AD structure
	FF # Manufacturer specific data AD type
	4C 00 # Company identifier code (0x004C == Apple)
	02 # Byte 0 of iBeacon advertisement indicator
	15 # Byte 1 of iBeacon advertisement indicator
	e2 c5 6d b5 df fb 48 d2 b0 60 d0 f5 a7 10 96 e0 # iBeacon proximity uuid
	00 00 # major 
	00 00 # minor 
	c5 # The 2's complement of the calibrated Tx Power

Is it possible to do that with the Android Beacon Library?

davidgyoung:
Unfortunately, this is not currently allowed by the Android BLE APIs, so the library cannot this.

PR #683, opened 2018-05-08
Adds connectable property to BeaconTransmitter

mharper:
Being able to set a beacon advertisement as "connectable" causes 3
additional bytes to be prepended to the advertisement PDU.

The value of these 3 bytes are: `0x02011a`.

They are:
1. PDU length (02)
2. PDU type (01)
3. Payload (1a)

These 3 bytes are present in an iBeacon advertisement on iOS. However,
they do not appear in the equivalent advertisement on Android using
`BeaconTransmitter` without setting the `AdvertiseSettings` to
"connectable."

Ref #141.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+### 2.13.2 / 2018-05-08
+
+[Full Changelog](https://github.com/AltBeacon/android-beacon-library/compare/2.13.2...2.13.1)
+
+Enhancements:
+ - BeaconTransmitter advertisements may be configured as connectable (#683, Michael Harper)
+
 ### 2.13.1 / 2018-03-05
 
 [Full Changelog](https://github.com/AltBeacon/android-beacon-library/compare/2.13.1...2.13)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,4 @@
-### 2.13.2 / 2018-05-08
-
-[Full Changelog](https://github.com/AltBeacon/android-beacon-library/compare/2.13.2...2.13.1)
+### Development
 
 Enhancements:
  - BeaconTransmitter advertisements may be configured as connectable (#683, Michael Harper)
--- a/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
+++ b/src/main/java/org/altbeacon/beacon/BeaconTransmitter.java
@@ -1,21 +1,21 @@
 package org.altbeacon.beacon;
 
-import org.altbeacon.beacon.logging.LogManager;
-
 import android.annotation.TargetApi;
 import android.bluetooth.BluetoothAdapter;
 import android.bluetooth.BluetoothManager;
-import java.nio.ByteOrder;
-import java.nio.ByteBuffer;
-import java.util.UUID;
 import android.bluetooth.le.AdvertiseCallback;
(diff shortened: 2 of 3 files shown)


davidgyoung:
Thanks, @mharper.  This begs the question, should the default be connectable?  Perhaps not for backward compatibility.  And just because iOS behaves that way  by default on transmission doesn't mean Android should -- I think it is a valid iBeacon packet either way.

mharper:
@davidhelms @syoder do you guys have anything to add?

mharper:
@davidgyoung please note that when I was running `./gradlew release`, the test `allowsAccessToRegionsAfterRestore` consistently failed with the message:
```
java.lang.AssertionError: beaconManager should return restored regions expected:<50> but was:<0>
```

The test was successful in isolation and when run along with all the tests in `MonitoringStatusTest`. But it consistently failed when run as part of running `release`. @cupakromer and I collaborated on the modification in this PR which fixes the problem.

davidgyoung:
I had recently noted that that test was failing, and looked at it briefly, but had not spent time to figure out why.  Perhaps it is ordering of tests that can cause it to fail, and some external update to the test runner ordering caused it to start failing.   That's just a guess.   Do you know anything more?

mharper:
@davidgyoung Is that change to `CHANGELOG.md` what you were looking for?

plottsjp:
@mharper @davidgyoung I have a reason why you would want to make the connectable property true by default: The Samsung s8 will not detect software created ibeacons without the advertisement flags being set. 

I found this issue because we have a app that that both detects and transmits as an iBeacon using the Altbeacon Library (Thanks for the library btw!). When we tested our app on a Samsung s8 we found that the app would only detect hardware ibeacons. It would not detect beacons simulated by our app or by another app ([BeaconSimulator](https://play.google.com/store/apps/details?id=net.alea.beaconsimulator&hl=en)). Once I set the connectable property to be true in my app the Samsung S8 was able to detect our simulated beacons again.

davidgyoung:
Thanks for the thoughts, @plottsjp.  I just did a quick test on a Galaxy S9 (admittedly not the S8 you mentioned) using a Locate app (based on an older version of this library without the connectable  setting)  transmitter, and had no trouble detecting it on the S9.   I've also done successful tests detecting non-connectable transmissions on the S5. 

Any reason to think this is an S8-specific issue?  Any other variables that you think may be at play in what you saw?

plottsjp:
I don't know if this is only for the S8 or if other phones do this as well. However the S8 is and will be one of the most popular android devices (in the USA) for a while, so it is probably worth fixing even if this is unique to the S8.

There shouldn't be any other variables in play. I noticed my s8 couldn't receive a phone based beacon and i found this thread. So i updated my app to use the latest version of the Altbeacon library (3.14). This did not fix the bug. Then I set connectable = true which fixed the bug. after these changes the s8 was able to detect beacons from the updated version of our app. (additionally i tried this on 2 different galaxy S8 devices and both of them responded the same way before and after the fix).

I reached out to the developer of the [BeaconSimulator](https://play.google.com/store/apps/details?id=net.alea.beaconsimulator&hl=en) and he said that he had other galaxy s8 user complaining about similar connection issues. (Though he is not using the Altbeacon library to configure his beacons).

PR #702, opened 2018-06-11
Cancel schedule jobs after last consumer unbound

### LINKED ISSUE #657: unbind does not stop scan jobs
### Expected behavior

calling `beaconManager.unbind(...)` will stop any ScanJobs that are running or that are scheduled.

### Actual behavior

The scheduled jobs keep running.

Do the following on Android 8+:

1. Call `beaconManager.bind(...);`
2. Call `beaconManager.startRangingBeaconsInRegion(...);`
3. Call `beaconManager.unbind(...);`

Note that you still get callbacks to `didRangeBeaconsInRegion` despite the call to unbind.

Android 8+

2.13.1

davidgyoung:
Before this change, calling BeaconManager.unbind(...) would never stop scheduled scanning jobs if they were in use.  This fixes that so after the last consumer is unbound, all scheduled scan jobs will be cancelled.

Fixes #657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Development
+
+Bug Fixes:
+ - Stop running scheduled jobs to do sscans after last consumer unbound. (#702, David G. Young)
+
 ### 2.15 / 2018-07-04
 
 Enhancements:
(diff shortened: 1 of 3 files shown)


himanshu-android:
This fix works for me. Tested on Nokia 6.1 running 8.1.0.

davidgyoung:
Thanks for the tests, @himanshu-android. 

I suspect the reason that most library users are not affected by this is that the most common ways to use the library is to bind only once on app startup, and never call `unbind()`.

kavin626:
How i can i stop these behaviour,i am getting this behaviour now on android 8.1.Please

davidgyoung:
@kavin626 you simply need to call beaconManager.unbind(...) on all consumers to get the jobs to stop.

Issue #638, opened 2018-01-12
Attempt to invoke virtual method 'boolean org.altbeacon.beacon.service.RangedBeacon.isTracked()' on a null object reference


Eider3a:
Hi, I have the next issue using the library AltBeacon, can you help me to fix it, sometimes I have this trouble (java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.altbeacon.beacon.service.RangedBeacon.isTracked()' on a null object reference), appers to be in the file RangeState.class in the line 59, I tried to modify the library but it is lock.

Thanks for all.
This is my code in MainActivity:

`
@Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);

beaconManager=BeaconManager.getInstanceForApplication(context);
        beaconManager.getBeaconParsers().add(new BeaconParser().
                setBeaconLayout("m:2-3=0215,i:4-19,i:20-21,i:22-23,p:24-24,d:25-25"));
        beaconManager.bind(this);

    }

    @Override
    public void onBeaconServiceConnect() {
        final Region region=new Region("MyBeacon",null,null,null);

        beaconManager.addRangeNotifier(new RangeNotifier() {
            //Se ejecuta cada segundo para verificar si existen beacons alrededor.
            @Override
            public void didRangeBeaconsInRegion(Collection<Beacon> beacons, Region region) {
                beaconList.clear();//Limpio la lista.

                for(Beacon oneBeacon:beacons){

                    final String dirMAC=oneBeacon.getBluetoothAddress();
                    final String uuid=oneBeacon.getId1().toString();
                    final String minor=oneBeacon.getId2().toString();
                    final String major=oneBeacon.getId3().toString();
                    final String distancia=String.valueOf(oneBeacon.getDistance());

                    //Crear la lista.
                    Beacons beaconElement=new Beacons(uuid,minor,major,dirMAC,distancia);
                    beaconList.add(beaconElement);

                    oneBeacon.setHardwareEqualityEnforced(true);

                    Log.i("MainActivity", "dist:"+oneBeacon.getDistance()+" adr:"+oneBeacon.getBluetoothAddress()+" txP"+oneBeacon.getRssi());

                }
                //shows the list in a recyclerview.
                runOnUiThread(new Runnable() {
                    public void run() {
                        beaconAdapter.notifyDataSetChanged();
                    }
                });

            }
        });

        try {
[...]

PR #716, opened 2018-08-02
Fix NPE on RangedBeacon per #638

davidgyoung:
This NPE probably happens in a race condition where the RangedBeacon is removed by another thread in the middle of the two step process to (1) check if the RangedBeacon is in the collection and (2) pull the RangedBeacon from the collection, causing a NPE in step 2.

This change fixes that problem by making it a one step process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### Development
 
 Bug Fixes:
- - Fix intermittent NPE on ranging beacons (#716, David G. Young)
+ - Fix intermittent NPE on ranging beacons (#716, Federico Bertoli, David G. Young)
 
 ### 2.15 / 2018-07-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Development
+
+Bug Fixes:
+ - Fix intermittent NPE on ranging beacons (#638, David G. Young)
+
 ### 2.15 / 2018-07-04
 
 Enhancements:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### Development
 
 Bug Fixes:
- - Fix intermittent NPE on ranging beacons (#638, David G. Young)
+ - Fix intermittent NPE on ranging beacons (#716, David G. Young)
 
 ### 2.15 / 2018-07-04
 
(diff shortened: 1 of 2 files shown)


davidgyoung:
Good catch, @thebertozz.  I just added this change to the pull request and gave you credit in the change log.

FalconBFR:
Dear all, this issue is still not solved in 2020, using the latest library. 
Specifically, when I call the function from anther file in the same directory (it is specifically a service class, but I tried a normal class and it didn't work either), it still creates this 'Null Pointer Error'. There is no problem when calling it within the file. Here is the method I called it:

**Before splitting it into two files, the function was (in main activity):**

startButton.setOnClickListener((v) -> { startBeaconMonitoring(); });

**No change to the function below:, for refeerence**
public void startBeaconMonitoring() {
        Log.d(TAG, "startBeaconMonitoring called");
        System.out.println("START BEACON MONITORING FUNCTION");
        try {

            //beaconRegion = new Region("MyBeaconStuff", null, null, null);
            //beaconRegion = new Region(null);
            beaconManager.startMonitoringBeaconsInRegion(beaconRegion);
            beaconManager.startRangingBeaconsInRegion(beaconRegion);
            transmitbeacon();
        } catch (RemoteException e) {
            e.printStackTrace();
        }
    }

**New Code:

Main Activity:** 
...
        startButton.setOnClickListener((v) -> {
            startService(new Intent (this, beaconcontrolservice.class));
        });
...

**In the file : beaconcontrolservice.java (same directory):**

package com.example.beaconwithoracletrainingvideo;

import android.app.Service;
import android.content.Intent;
import android.os.IBinder;

import androidx.annotation.Nullable;

public class beaconcontrolservice extends Service {
    @Nullable
    @Override
    public IBinder onBind(Intent intent) {
        return null;
    }

    @Override
    public int onStartCommand(Intent intent, int flags, int startId) {
        MainActivity MainActivity = new MainActivity();
        //return super.onStartCommand(intent, flags, startId);
        MainActivity.startBeaconMonitoring();
        return START_STICKY;
    }

    @Override
    public void onDestroy() {
        super.onDestroy();
    }
}

**Back to Main activity**

    public void startBeaconMonitoring() {
        Log.d(TAG, "startBeaconMonitoring called");
        System.out.println("START BEACON MONITORING FUNCTION");
        try {
            //beaconRegion = new Region("MyBeacons", Identifier.parse("0AC59CA4-DFA6-442C-8C65-22247851344C"),
              //      Identifier.parse("4"),Identifier.parse("200"));
            //beaconRegion = new Region("MyBeaconStuff", null, null, null);
            //beaconRegion = new Region(null);
            beaconManager.startMonitoringBeaconsInRegion(beaconRegion);
            beaconManager.startRangingBeaconsInRegion(beaconRegion);
            transmitbeacon();
        } catch (RemoteException e) {
            e.printStackTrace();
        }
    }

PR #748, opened 2018-10-04
Fix leaked NotificationProcessor in ScanJobsScheduler

### LINKED ISSUE #735: didRangeBeaconsInRegion() called multiple times in foreground mode
### Expected behavior

Using the foreground mode, `didRangeBeaconsInRegion()` should be called once for every scan action.

### Actual behavior

The first time I call `startRangingBeaconsInRegion()` it works as expected. If I stop and start again, the listener is called twice. In version 2.15 it works.

We are using the Foreground Service Scanning.

The first time I call `startRangingBeaconsInRegion()` it works as expected. 

Then, if a call `stopRangingBeaconsInRegion()` and `startRangingBeaconsInRegion()`, the method `didRangeBeaconsInRegion()` is called twice for every scan. I'm calling `addRangeNotifier()` every time before  `startRangingBeaconsInRegion()`, but this is working with previous versions of the library.

If a do this a third time, the listener is called three times.

Google Pixel XL, Android 9
Xiaomi Redmi 3S, Android 6.0.1

2.15.1

adrinieto:
This caused multiple `NotificationProcessor` registrations when binding and unbinding.

This PR and #745 will completely fix #735


davidgyoung:
Thanks, @adrinieto

Issue #758, opened 2018-10-25
 'void org.altbeacon.beacon.service.scanner.CycledLeScanner.stop()' on a null object reference


kababu007:
### Expected behavior
Should not crash when the library stops scanning for beacons.

### Actual behavior
Crash recorded:
java.lang.NullPointerException: Attempt to invoke virtual method 'void org.altbeacon.beacon.service.scanner.CycledLeScanner.stop()' on a null object reference
	at org.altbeacon.beacon.service.ScanJob.stopScanning()(unknown:169)
	at org.altbeacon.beacon.service.ScanJob.onStopJob()(unknown:161)
	at android.app.job.JobService$1.onStopJob()(JobService.java:76)
	at android.app.job.JobServiceEngine$JobHandler.handleMessage()(JobServiceEngine.java:117)
	at android.os.Handler.dispatchMessage()(Handler.java:106)
	at android.os.Looper.loop()(Looper.java:164)
	at android.app.ActivityThread.main()(ActivityThread.java:6651)
	at java.lang.reflect.Method.invoke()(Method.java:-2)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run()(RuntimeInit.java:438)
	at com.android.internal.os.ZygoteInit.main()(ZygoteInit.java:810)

### Steps to reproduce this behavior
Not able to replicate

### Mobile device model and OS version
Device Model  LG-Q710AL
Device Name   LGE
Device Type    LG Device
OS Version     8.1.0

### Android Beacon Library version
2.15.2

IMPORTANT:  This forum is reserved for feature requests or reproducible bugs with the library itself.  If you need help with using the library with your project, please open a new question on StackOverflow.com.

kababu007:
Just verified the crash logs and it happens only with version 2.15.2

ryot:
One of our partners is also reporting this crash after updating to 2.15.2, sourced from a OnePlus One running Android 8.1
```
java.lang.RuntimeException:
at android.app.job.JobServiceEngine$JobHandler.handleMessage (JobServiceEngine.java:121)
at android.os.Handler.dispatchMessage (Handler.java:106)
at android.os.Looper.loop (Looper.java:164)
at android.app.ActivityThread.main (ActivityThread.java:6494)
at java.lang.reflect.Method.invoke (Method.java)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:440)
[...]

PR #724, opened 2018-08-15
Catch of exception thrown by Android stopping scan

davidgyoung:
This is an attempt to fix #701.   I have not tested that this works, as there are no known ways to reproduce the problem.  It has only been seen in crash logs.

PR #766, opened 2018-11-06
Fix NPE on cycled le scanner

davidgyoung:
Fixes  crash issue reported in #758.  This is a side effect of the attempted  fix in #724.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Development
+
+- Prevent NPE on access CycledLEScanner after OOM on Android 8+.  (#766, David G. Young)
+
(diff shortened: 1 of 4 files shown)


vitas:
I'm not sure I understand the change is made here
        if (mScanHelper.getCycledScanner() == null) {
            mScanHelper.getCycledScanner().stop();
            mScanHelper.getCycledScanner().destroy();
        }
I believe it will throw NPE

davidgyoung:
Indeed, the logic was accidentally reversed.  That was fixed in #804

vitas:
ok. sorry i did know it was fixed. thank you

Issue #212, opened 2015-06-10
NullPointerException in BeaconService.matchingRegions()


mykolaj:
Hello!
I've got an issue which doesn't seem to depend on Android version or AltBeaconLib version. Anyways, it is currently confirmed to occur on some of Kyocera and Samsung devices running Android 4.4 and Android 5.0 and AltBeaconLib 2.1.4. Here's a stack trace:

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean org.altbeacon.beacon.Region.matchesBeacon(org.altbeacon.beacon.Beacon)' on a null object reference
at org.altbeacon.beacon.service.BeaconService.matchingRegions(BeaconService.java:462)
at org.altbeacon.beacon.service.BeaconService.processBeaconFromScan(BeaconService.java:399)
at org.altbeacon.beacon.service.BeaconService.access$400(BeaconService.java:66)
at org.altbeacon.beacon.service.BeaconService$ScanProcessor.doInBackground(BeaconService.java:439)
at org.altbeacon.beacon.service.BeaconService$ScanProcessor.doInBackground(BeaconService.java:422)
at android.os.AsyncTask$2.call(AsyncTask.java:288)
at java.util.concurrent.FutureTask.run(FutureTask.java:237)
... 3 more
```

So far it happens from time to time without any visible regularity, so i can't reproduce it on purpose, but it happens quite often. Here's some code details:
`// Start ranging`
`mRegion = new Region(UUID.randomUUID().toString(), null, null, null);`
`mBeaconManager.startRangingBeaconsInRegion(mRegion);`
 (UUID is changed when a program is restarted)

After some time (could be minutes, could be hours) a program crashes with the above exception.
Also it's worth to say that scan is searching for iBeacons, and seams to happen when there are a lot of Bluetooth devices nearby .

This exception happens inside next loop: 

``` java
           Iterator<Region> regionIterator = regions.iterator();
            while (regionIterator.hasNext()) {
                Region region = regionIterator.next();
                if (region.matchesBeacon(beacon)) {
                    matched.add(region);
                } else {
                    LogManager.d(TAG, "This region (%s) does not match beacon: %s", region, beacon);
                }
            }
```

Of course a quickfix would be to just check if a Region is null and continue the loop if it is. Or is there any purpose for `null` key in `rangedRegionState` and `monitoredRegionState` maps?
Thank you!

davidgyoung:
[...]

PR #770, opened 2018-11-10
Protect against a NPE on iteration of regions

davidgyoung:
This is a reimplementation of #215, whose changes were inadvertently lost when the ScanHelper was refactored.

This protects against a NullPointerException when iterating regions for a beacon match, presumably because the list of regions changed during the iteration.


davidgyoung:
This fixes #212 (again)

PR #799, opened 2018-12-11
Crash fix: bluetoothCrashResolver is used only in versions before Lollipop

mikkoau:
BeaconService shutdown causes a crash on newer Androids because it tried to stop bluetoothCrashResolver which is only initialized in Android versions before Lollipop.


davidgyoung:
Thanks for this, @mikkoau.   I believe this is already covered by this change #790.  You probably did not noticed it because it has not been put into a release yet.  I expect that release to go out imminently.

PR #803, opened 2018-12-19
Sample

TonyTangAndroid:


davidgyoung:
Wow, thanks for the big effort here, @TonyTang. 

There are a lot of good changes here, but boy it is huge!  The trouble is that there is so much in a single pull request that it makes it super hard to review everything to make sure there are not problems or unintended consequences.

Any chance you could break this up into several separate PRs that are more easily digested?   Perhaps:

* Robolectric upgrade
* Gradle cleanup
* Combining ref app with main repo (why do you think this is better)
* Gitignore cleanup
* Reformat code

The above would make this much less effort to properly review and get merged.

TonyTangAndroid:
This pull request also comes as a surprise to me because I totally did not aware of such action. Obviously, this is a pull request that is too huge to be merged. And I was not even set out to get my code merged here because I realize that it is highly likely to introduce changes that is not within your plan. But I am very glad to know that you are actually open for such potential changes. Hence, I would like to give it a proper conduct steps by step in per your suggestion. And before this, there are a few items that I would like to discuss here: 

Robolectric Upgrade to 4.1 will go together with bumping target SDK to 28 and android support library to 28.0.0. 
This sounds good to you?

Combining ref app makes it easy for us to verify our library's changes without going through another repository. And it is a common practice. Almost all open source project includes the minimum sample in its the same repository. So anyone else who is interested in making contribution will immediately know that he or she could start from running the sample app to get the hand dirty before diving deeper. Furthermore, there had been several occasion that I had to make an effort to locate the reference app because the reference app is not included in the library. And the reference app strikes to users that it is not actively maintained and it is NOT a proper demonstration of how to use the library in PROD way. With all being said, it will be good to include it here. 


I am not used to the fact that the file gets changed randomly simply because I open the file. It will resolve such issues by reformatting code using default Android studio settings and this will also set a base for other people' contribution.

TonyTangAndroid:
I think I know the reason why the pull request is here. I was intended to send the pull request to my forked master branch. Sorry for the confusion.

davidgyoung:
No problem on the accidental pull request.  I will close it so as not to cause confusion.

Sure, your plan with the robolectric, tools and sdk upgrades sounds fine.  And, yes, I see the advantages to combining the reference app, too.

PR #805, opened 2018-12-20
Bump version Gradle to 4.10.3 and Robolectric to 4.1 

TonyTangAndroid:
Bump version Gradle to 4.10.3 and Robolectric to 4.1.

All unit tests are passed. The logger issue is actually rooted from ASOP, which pollutes our test cases. 
I updated the assertion to make it pass. 

This pull request construes the minimum changes to support the latest gradle, Android and Robolectric and is ready to be merged.

I will make another pull request to optimize the build script.


TonyTangAndroid:
I have more following up pull request to add up. But I would like to hold on until we resolve this pull request.

TonyTangAndroid:
If you run the failed test independently, it will pass. If you run all the test all together, it will fail. If you run to the failed test continuously for 10 times, it will only pass for the first time. I tried to find the cause, but to no prevail.

davidgyoung:
I was able to fix the intermittently failing unit tests by more clearly setting the initial conditions in the `before` method, and consistently accessing the same object in all tests.

TonyTangAndroid:
Awesome! That is very good to know. I would like to make more pull request related to sample code.  We could discuss more in the corresponding thread.

davidgyoung:
sounds good, @TonyTangAndroid

PR #811, opened 2019-01-10
Organize the build script

TonyTangAndroid:
Organize the build script to make it as a standardized Android library.


TonyTangAndroid:
This might be a large pull request. It does two things : 
1, organize the build script.
2, organize the file trees to make it standardized.


Let me know if you want to break it into two pull requests.

The next step is to format the code base based on default Android Studio settings and introduce the sample code, etc.

davidgyoung:
I like the step by step approach, @TonyTangAndroid, thanks.  Testing this out now.

So basically the main changes are to:
* move the library source into /lib
* break up the build.gradle file

Anything else I am missing?

TonyTangAndroid:
Yes, that's all. Cool. I am to break it down and send two pull request. We could close this one.

davidgyoung:
Nah, that's fine.  I think this one is small enough.

It looks like the gradle changes may change some of the target usage.  For example, running `./gradlew test` no longer runs the tests.

If any of these change, can you please update the new ones in the README.md?

davidgyoung:
Ah, wait, I guess they do run.  The reports are now in lib/build/reports/tests/, which is as should be expected.  Can you please glance over the other targets in the README and see if you see anything else that changes?

TonyTangAndroid:
Cool. Right away.

davidgyoung:
I tested each target, and all look fine.  I'm a little surprised nothing broke!  I think this is good to merge unless you see anything else.

TonyTangAndroid:
Cool. I just updated the version that is applied to verify the build. I am all good to go here.

Issue #817, opened 2019-02-04
RegionBootstrap constructor parameters


tekinalper:
https://github.com/AltBeacon/android-beacon-library/blob/b9876b45acd6c72c84b8d0325062259020eb89ba/lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java#L49

As a SDK developer, I get Application context from App and implement BootstrapNotifier in the SDK. With the current constructors, I'm not able to set the two different parameter in one. 
I found a solution by creating a custom RegionBootstrap. In case of someone needs such a solution like I did, you may add new constructors or I can send PR if it makes sense for you.  

`public RegionBootstrap(BootstrapNotifier bootstrapNotifier, Application application, Region region) {`

`public RegionBootstrap(BootstrapNotifier bootstrapNotifier, Application application, List<Region> regions) {`

davidgyoung:
@tekinalper, I agree it would be useful to more easily separate the context from the notifier.  Is there a way to make this backward-compatible?

It seems such a solution would have to redefine BootstrapNotifier so it no longer has a getApplicatonContext method.  

Thoughts?

tekinalper:
Yes, it was the another issue! Currently, I don't initialize RegionBootstrap if the application context is null. But also, setting application context (assuming nonnull) to getApplicationContext method makes me feel not so comfortable. 

It seems it won't be really backward-compatible because of the interface. Making getApplicatonContext optional is not possible. 

How about removing it? :)

davidgyoung:
How about making the method signature look like this:

`public RegionBootstrap(Context context, MonitorNotifier monitorNotifier, Region region)`

Then internally you could create a new RegionBootstrap by constructing a new BootstrapNotifier as an anonymous inner class made up of the passed monitorNotifier and context with chained methods.  Something like:

```
mContext = context;
mMonitorNotifier = monitorNotifier;
mBootstrapNotifier = new BootstrapNotifier() {
  public Context getApplicationContext() {
    return mContext.getApplicationContext();
  }
  public void didEnterRegion(Region region) {
    return mMonitorNotifier.didEnterRegion(region);
  }
  ...
  
}
```

tekinalper:
Looks much better 👍

davidgyoung:
OK, cool.  Do you maybe want to play with an implementation like that in your code, make sure it works OK, then issue a PR?

davidgyoung:
Fixed in #821, released in 2.16.1

PR #821, opened 2019-02-05
Change RegionBootstrap method signature and separate the context from…

tekinalper:
… the notifier

#817

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
+++ b/lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
@@ -46,14 +46,14 @@ public class RegionBootstrap {
      * Constructor to bootstrap your Application on an entry/exit from a single region.
      *
      * @param context
+     * @param monitorNotifier
      * @param region
      */
     public RegionBootstrap(final Context context, final MonitorNotifier monitorNotifier, Region region) {
-        if (context.getApplicationContext() == null) {
+        if (context == null) {
             throw new NullPointerException("Application Context should not be null");
         }
-        beaconManager = BeaconManager.getInstanceForApplication(context.getApplicationContext());
-        this.context = context;
+        this.context = context.getApplicationContext();
         this.monitorNotifier = monitorNotifier;
 
         this.monitorNotifier = new BootstrapNotifier() {
@@ -76,6 +76,8 @@ public class RegionBootstrap {
 
         regions = new ArrayList<Region>();
         regions.add(region);
+
+        beaconManager = BeaconManager.getInstanceForApplication(context);
         beaconConsumer = new InternalBeaconConsumer();
         beaconManager.bind(beaconConsumer);
         LogManager.d(TAG, "Waiting for BeaconService connection");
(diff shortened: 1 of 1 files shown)


lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -34,24 +35,106 @@
...
     private List<Region> regions;
     private boolean disabled = false;
     private BeaconConsumer beaconConsumer;
     private boolean serviceConnected = false;
 
+    /**
+     * Constructor to bootstrap your Application on an entry/exit from a single region.
+     *
+     * @param context
+     * @param region
```
davidgyoung: javadoc comment should list the new monitorNotifier parameter

lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -34,24 +35,106 @@
...
+        beaconManager.bind(beaconConsumer);
+        LogManager.d(TAG, "Waiting for BeaconService connection");
+    }
+
+    /**
+     * Constructor to bootstrap your Application on an entry/exit from multiple regions
+     *
+     * @param context
+     * @param regions
+     */
```
davidgyoung: javadoc comment should list the new monitorNotifier parameter

lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -34,24 +35,106 @@
...
+    public RegionBootstrap(final Context context, final MonitorNotifier monitorNotifier, List<Region> regions) {
+        if (context.getApplicationContext() == null) {
```
davidgyoung: might be a good idea to just check for context == null, since that is the more likely case

lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -34,24 +35,106 @@
...
+        beaconManager.bind(beaconConsumer);
+        LogManager.d(TAG, "Waiting for BeaconService connection");
+    }
+
     /**
      * Constructor to bootstrap your Application on an entry/exit from a single region.
      *
      * @param application
      * @param region
+     * @deprecated This will be removed in a later release. Use
```
davidgyoung: Is there any reason to deprecate this?  It seems like it is a simpler way of using RegionBootstrap if you are using the Application class as the place to get your callbacks.

lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
```diff
@@ -34,24 +35,106 @@
...
+     * * {@link org.altbeacon.beacon.startup.RegionBootstrap#RegionBootstrap(Context, MonitorNotifier, Region)}}
+     * * instead.
      */
     public RegionBootstrap(BootstrapNotifier application, Region region) {
         if (application.getApplicationContext() == null) {
             throw new NullPointerException("The BootstrapNotifier instance is returning null from its getApplicationContext() method.  Have you implemented this method?");
         }
         beaconManager = BeaconManager.getInstanceForApplication(application.getApplicationContext());
-        this.application = application;
+        this.context = application.getApplicationContext();
```
davidgyoung: if we consistently set this.context = application.getApplicationContext() in all constructors (which this appears to do), then all other places where we call this.context.getApplicationContext() can be changed to just this.context

lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
tekinalper: Just thought that it may cause confusion but yes why not use BootstrapNotifier if you only have Application class.

Issue #636, opened 2018-01-02
NullPointerException - ScanHelper class


himanshu-android:
### Expected behavior
Start background scanning in Android Oreo

### Actual behavior
Application crashed due to NullPointerException

### Steps to reproduce this behavior
Unavailability of Android O device but crash report should help.

### Mobile device model and OS version
OS Version: 8.0.0
Device: ONEPLUS A3003

### Android Beacon Library version
Version - 2.12.4

[le_scan_exception.txt](https://github.com/AltBeacon/android-beacon-library/files/1597417/le_scan_exception.txt)

ncapdevi:
I am getting a similar crash

```
Fatal Exception: java.lang.RuntimeException: java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Set com.android.bluetooth.gatt.ScanManager.getBatchScanQueue()' on a null object reference
       at android.app.job.JobServiceEngine$JobHandler.handleMessage(JobServiceEngine.java:112)
       at android.os.Handler.dispatchMessage(Handler.java:105)
       at android.os.Looper.loop(Looper.java:164)
       at android.app.ActivityThread.main(ActivityThread.java:6809)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.Set com.android.bluetooth.gatt.ScanManager.getBatchScanQueue()' on a null object reference
       at android.os.Parcel.readException(Parcel.java:1948)
       at android.os.Parcel.readException(Parcel.java:1888)
       at android.bluetooth.IBluetoothGatt$Stub$Proxy.stopScanForIntent(IBluetoothGatt.java:970)
       at android.bluetooth.le.BluetoothLeScanner.stopScan(BluetoothLeScanner.java:290)
       at org.altbeacon.beacon.service.ScanHelper.stopAndroidOBackgroundScan(ScanHelper.java:194)
       at org.altbeacon.beacon.service.ScanJob.restartScanning(ScanJob.java:183)
       at org.altbeacon.beacon.service.ScanJob.startScanning(ScanJob.java:221)
       at org.altbeacon.beacon.service.ScanJob.onStartJob(ScanJob.java:77)
[...]

PR #824, opened 2019-02-08
ScanHelper Moving NPE exception above RuntimeException

kababu007:
ScanHelper Moving NPE exception above RuntimeException


kababu007:
ScanHelper Moving NPE exception above RuntimeException

kababu007:
Updated the changelog as below:

### 2019-02-09 (#636, kababu007)
- Fix intermittent crash caused by internal Android NPE

PR #869, opened 2019-04-30
WIP: Use hardware filters for faster exit detections

davidgyoung:
This adds support for region-specific hardware filters to give you quick region exit events on pattern loss.  Exit events are often detected in about 20 seconds, but testing on the Galaxy S9 this is not 100% reliable.

For this to work, you must monitor at most one region with at least one identifier.  If these conditions are not met, hardware filters are not used to accelerate exit events.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'com.android.tools.build:gradle:3.3.2'
         //noinspection GradleDependency
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.1'
         //noinspection GradleDependency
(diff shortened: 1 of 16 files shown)


davidgyoung:
In testing on a Samsung Galaxy S9, I found that the pattern lock loss callback from a scan filter is not 100% reliable.  Here's why:

1. If you turn on the scan running with pattern loss filter only after a pattern was detected (after a first detection scan returns a result) then you do not always get a pattern loss callback.  I suspect this happens because for the pattern loss callback to fire the hardware must detect at least one packet before the pattern is lost.    It is unclear how long it will take for the hardware to determine it has seen the pattern.  This indeterminate problem might be fixed by leaving it on all the time, but this causes problem (2).
 
2. If you always keep a scan running with pattern loss filter, you get repeated callbacks telling you the pattern detection has been lost.  This happens even though there was no new detection since the last callback telling you the pattern was lost.  This keeps waking up the app and burns battery.  This alternative is a non-starter.

Even with limitation from (1), the code in this branch *often* accelerates region exit events, but it does not work 100% of the time, meaning that it sometimes takes until the next full software scan to detect a region exit.

davidgyoung:
I am closing this PR as testing shows this is just not workable

PR #882, opened 2019-05-23
[android-beacon-library] Fix #881 Background Monitoring using RegionB…

### LINKED ISSUE #881: Background Monitoring using RegionBootstrap is not calling didExitRegion
While monitoring a region in the Background and leaving the region after it's been detected, didExitRegion is not triggered because the scanning gets stopped when the region became in range.
### Expected behavior
Scanning should keep running even when a region is in range so that other regions and leaving the range can still be detected
### Actual behavior
Scanning stops once a range is detected
- Start the reference app
- Send the app to background
- Enter beacon range
<< Scanning stops>> 
- Leave the range
<<didExitRegion doesn't get triggered>> 
- Pixel 3 running Android Q-beta3
- Pixel 2 running Android P
2.16.1
IMPORTANT:  This forum is reserved for feature requests or reproducible bugs with the library itself.  If you need help with using the library with your project, please open a new question on StackOverflow.com.

mohand1993:
…ootstrap is not calling didExitRegion

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/service/ScanJob.java
+++ b/lib/src/main/java/org/altbeacon/beacon/service/ScanJob.java
@@ -136,15 +136,12 @@ public class ScanJob extends JobService {
         }
         if (insideAnyRegion) {
             // TODO: Set up a scan filter for not detecting a beacon pattern
-            LogManager.i(TAG, "We are inside a beacon region.  We will not scan between cycles.");
+        }
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            mScanHelper.startAndroidOBackgroundScan(mScanState.getBeaconParsers());
         }
         else {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                mScanHelper.startAndroidOBackgroundScan(mScanState.getBeaconParsers());
-            }
-            else {
-                LogManager.d(TAG, "This is not Android O.  No scanning between cycles when using ScanJob");
-            }
+            LogManager.d(TAG, "This is not Android O.  No scanning between cycles when using ScanJob");
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/src/main/java/org/altbeacon/beacon/service/ScanJob.java
```diff
@@ -136,15 +136,12 @@ private void startPassiveScanIfNeeded() {
         }
         if (insideAnyRegion) {
             // TODO: Set up a scan filter for not detecting a beacon pattern
-            LogManager.i(TAG, "We are inside a beacon region.  We will not scan between cycles.");
+        }
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            mScanHelper.startAndroidOBackgroundScan(mScanState.getBeaconParsers());
```
davidgyoung: I don't think this code change will be helpful.  If you are already inside a beacon region (meaning the beacon is visible in the vicinity), there is no point in calling `mScanHelper.startAndroidOBackgroundScan(mScanState.getBeaconParsers());`.  This will just result in callbacks quickly each time that an already visible beacon continues to be seen.  This will accomplish nothing more than using extra CPU and battery when the app is in the background.

davidgyoung:
@mohand1993, I have done a lot of research on the subject of fast region exit detections in #869.   Please read my description there and comment if you have any other ideas.

davidgyoung:
The change shown attempts to get faster region exits by setting up a filtered background scan looking for the beacon.  However, this won't help, because that filtered scan is looking for the beacon to be detected when it already is detected.

What would need to happen instead for this to work is to do a filtered scan for the beacon to disappear.  This is also called a "match loss" scan.  Unfortunately, there are obstacles imposed by the Android OS that make these impractical to use.  See #869 for a discussion.

mohand1993:
@davidgyoung 
Thanks for this explanation, I will have a look at the work you've done in #869  and will try to do some research myself. I will let you know if I find anything.

davidgyoung:
OK, @mohand1993.  Please do comment on #869 if you have other ideas.  I spent a large number of hours over the last couple of months trying to find a way to get fast region exits, but have hit some brick walls.  If you have other ideas, I'd be interested to hear them.

PR #893, opened 2019-06-05
catch NPE in BluetoothMedic#runScanTest

VanHoangTran:


davidgyoung:
Looks good, @VanHoangTran.  Thank you.

PR #921, opened 2019-10-01
Use bootstrap background with range notifier

rlatapy-luna:
Hi @davidgyoung 
I don't think you want to merge this, or at least not as it but could you take a look just to confirm there is no dangers to do that, especially regarding the energy consumption?

In a nutshell I'm just ranging instead of monitoring with the `RegionBootstrap` helper. From my first tests it seems to work well (coupled with a time check to avoid flood detection) and it solves our problem easily.

Thank you

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/BleNotifier.java
+++ b/lib/src/main/java/org/altbeacon/beacon/BleNotifier.java
@@ -1,0 +1,4 @@
+package org.altbeacon.beacon;
+
+public interface BleNotifier {
+}
--- a/lib/src/main/java/org/altbeacon/beacon/MonitorNotifier.java
+++ b/lib/src/main/java/org/altbeacon/beacon/MonitorNotifier.java
@@ -32,7 +32,7 @@ package org.altbeacon.beacon;
  *
  * @author David G. Young
  */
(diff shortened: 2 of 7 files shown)


davidgyoung:
@rlatapy-luna, I do not think there is any significant battery impact to his change.  The change effectively modifies RegionBootstrap so that you can use it to start Ranging instead of just Monitoring.  This should work fine.

That said, there really is an easier way to do this that doesn't require modifying the library.  You can simply use the RegionBootstrap as-is, then add code like this to start ranging immediately after setting up RegionBootstrap:

```
    @Override
    public void didDetermineStateForRegion(int state, Region region) {
        logToDisplay("Current region state is: " + (state == 1 ? "INSIDE" : "OUTSIDE ("+state+")"));
        BeaconManager beaconManager = BeaconManager.getInstanceForApplication(this);
        beaconManager.addRangeNotifier(this);
        beaconManager.startRangingBeaconsInRegion(region);
    }
```

The result will be that both monitoring and ranging will be set up at application startup instead of just one.  The effect will be otherwise identical to the code you have shown.

rlatapy-luna:
@davidgyoung Thank you, it's easier indeed

PR #931, opened 2019-11-07
Don't force bluetooth on, don't reset bluetooth when devices are paired

t0k4rt:
Hello, 
While I was inspecting BluetoothMedic I was wondering if some specific cases where taken into account.
I was specifically thinking about these two cases :
* Bluetooth was off: using cycleBluetooth would enable bluetooth while it was previously off
* BT Headphones (or other devices ?) where bonded: using cycleBluetooth would break bt connection with your bonded devices.

I think that would lead to bad user experience: 
* my bt was off but using this apps turns my bt on automatically
* I was listening to music it stops while using this app

I'm not really sure these cases may happen: if bt is in bad state does it still plays music on you headphones ?

Maybe we should prevent these edge case with minor checks just in case.
What do you think about this ?

### DIFF
--- a/lib/src/main/java/org/altbeacon/bluetooth/BluetoothMedic.java
+++ b/lib/src/main/java/org/altbeacon/bluetooth/BluetoothMedic.java
@@ -398,7 +398,13 @@ public class BluetoothMedic {
     private void cycleBluetooth() {
         LogManager.d(TAG, "Power cycling bluetooth");
         LogManager.d(TAG, "Turning Bluetooth off.");
-        if (mAdapter != null) {
+        // we do not force bt on
+        // we do not reset bluetooth while bonded with a device (you could deactivate bt headphones)
+        if (mAdapter != null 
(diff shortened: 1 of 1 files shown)


davidgyoung:
Thanks for the PR, @t0k4rt.  I certainly agree with the part about bluetooth being off.  Do you know that this actually happens?  I would think the bluetooth failure events that would trigger a power cycle would never be detected if bluetooth is off.  Have you seen this or do you just want to protect against it?

Correct me if I am wrong, won't getBondedDevices() return a list of all paired bluetooth devices, even if they are not in use?  If so, wouldn't this effectively disable the

t0k4rt:
@davidgyoung 
For the first point I haven't actually seen it, I just started to use the Medic in my code. But I Will dig deeper and try to check this.
For the second point, that's a good point ! I need to check wether the device is active or not. I'll check wether it's possible or not.

Issue #933, opened 2019-12-03
Background Scans on Android9 Samsung Devices


camaira:
Hello and first of all, thank you all very much for this library.

### Expected behavior
I use a foreground service for scanning for beacons, with a notification. I am using BLE Beacons as a Geofence. Fortunately I have many devices to test. The application works perfectly fine with Pixel3 (Android Q), Xiaomi Mi9SE (Android P), Oneplus 7 (Android P), Xiaomi Pocophone F1 (Android P), Motorola G6 (Android O). 

### Actual behavior
The beacon scan on Samsung Galaxy S10 and Galaxy Note 10 works fine if the screen is on. However the scan stops immediately if I turn the screen off, only to resume if it turns on again, even if the device is locked. Those two are the only phones where I see this behaviour.

### Steps to reproduce this behavior
The same problem occurs with the beacon reference application. So by uncommenting the code for the foregroundservice, you can reproduce this issue using a Galaxy S10 or Note 10 and trying to turn the screen off - the phone won't receive beacons any more.

setDebug(true) shows the following when in background:

`D/IntentHandler: got ranging data
D/CycledLeScanner: Waiting to stop scan cycle for another 97 milliseconds
D/CycledLeScanner: Done with scan cycle
D/ScanHelper: Beacon simulator not enabled
D/ScanHelper: Calling ranging callback
D/Callback: attempting callback via local broadcast intent: org.altbeacon.beacon.range_notification
D/CycledLeScanner: Not stopping scanning.  Device capable of multiple indistinct detections per scan.
    starting a new scan cycle
    We are already scanning and have been for 54231 millis
D/CycledLeScanner: Waiting to stop scan cycle for another 1100 milliseconds
    Scan started`

### Mobile device model and OS version
Test device is a Samsung Galaxy Note 10, Android 9.

### Android Beacon Library version
Latest, 2.16.3

I highly appreciate any help. Thank you in advance!

davidgyoung:
@camaira, Samsung made a custom change requiring that any beacon scan with the screen off have a non-empty scan filter.  The library was modified to accommodate this change, but for it to work the beacon you are using must match the scan filter the library generates.  You can read more about this here:

https://github.com/AltBeacon/android-beacon-library/issues/769

I suspect the beacon you are using may not be matching the Samsung-only scan filter used in the background.  This might happen if the beacon is uses a manufacturer advertisement (iBeacon, AltBeacon) but has a non-standard manufacturer identifier.

[...]

Issue #848, opened 2019-03-28
[Feature Request] Allow 128-bit service UUID for scan filters


mannodermaus:
### Android Beacon Library version
2.16.1

---

Currently, [`ScanFilterUtils#createScanFiltersForBeaconParsers()`](https://github.com/AltBeacon/android-beacon-library/blob/b9876b45acd6c72c84b8d0325062259020eb89ba/lib/src/main/java/org/altbeacon/beacon/service/scanner/ScanFilterUtils.java#L79-L90) expects a `BeaconParser`'s service UUID to be 16-bit - 128-bit UUID aren't supported. Our custom parser is being neglected on Samsung 8.1+ devices because of the non-empty scan filters applied on these devices (working around #769, also ref. #691)*. There's also the option to include manufacturer data in the scan filter (the `else` branch of the linked code above), however I wanted to explore the service UUID route first.

Is there a specific reason to only allow 16-bit UUIDs for scan filters, or would it be possible to have custom `BeaconParser` implementations expose a 128-bit service UUID as well? I realize this is a niche use case, and we might be able to work with the manufacturer data as well - however, UUID would be preferable in our case.

*Edit: Some more info on this. If we utilize a modified version of the library with the following change to `CycledLeScannerForLollipop`, our foreground scan works just fine:

```diff
- if (Build.MANUFACTURER.equalsIgnoreCase("samsung")) {
+ if (false && Build.MANUFACTURER.equalsIgnoreCase("samsung")) {
    // On the Samsung Galaxy Note 8.1, scans are blocked with screen off when the
    // scan filter is empty (wildcard).  We do a more detailed filter on Samsung only
    // because it might block detections of AltBeacon packets with non-standard
    // manufacturer codes.  See #769 for details.
    LogManager.d(TAG, "Using a non-empty scan filter since this is Samsung 8.1+");
    filters = new ScanFilterUtils().createScanFiltersForBeaconParsers(
            mBeaconManager.getBeaconParsers());
}
else {
    LogManager.d(TAG, "Using an empty scan filter since this is 8.1+ on Non-Samsung");
    // The wildcard filter matches everything.
    filters = new ScanFilterUtils().createWildcardScanFilters();
}
```

[...]

PR #936, opened 2019-12-13
Samsung devices stop scanning when screen is off

atotalnoob:
We have been encountering an issue where Android 9 Samsung devices are scanning with the screen is on, with wildcard scanning, then once the screen is off it doesn't switch over to non-wildcard scanning immediately, causing gaps in scanning to take place. 

Because the current code checks if it is interactive, it doesn't fall into non-wildcard scanning until the screen is off. By immediately going into non-wildcard scanning, we prevent gaps during the transition period. 

We were also able to create a work around the stops ranging/starts ranging for regions when the screen turns off

Possibly related to #914 #933

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScannerForLollipop.java
+++ b/lib/src/main/java/org/altbeacon/beacon/service/scanner/CycledLeScannerForLollipop.java
@@ -189,7 +189,7 @@ public class CycledLeScannerForLollipop extends CycledLeScanner {
             // We only add these filters on 8.1+ devices, because adding scan filters has been reported
             // to cause scan failures on some Samsung devices with Android 5.x
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O_MR1) {
-                if (Build.MANUFACTURER.equalsIgnoreCase("samsung") && !mPowerManager.isInteractive()) {
+                if (Build.MANUFACTURER.equalsIgnoreCase("samsung")) {
                     // On the Samsung Galaxy Note 8.1, scans are blocked with screen off when the
                     // scan filter is empty (wildcard).  We do a more detailed filter on Samsung only
                     // because it might block detections of AltBeacon packets with non-standard


atotalnoob:
It seems, as @davidgyoung fear, this did cause issues for existing usage. 

https://github.com/AltBeacon/android-beacon-library/issues/848#issuecomment-480614172

paulpv:
FYI, I suspect this is not actually fixed:
https://stackoverflow.com/a/57654627/252308

> I faced the same issue. I had Scan filters in order to scan BLE devices even if the screen were locked. But on Samsung devices it didn't work, so I search on Samsung forum and I discovered Knox SDK (https://seap.samsung.com/sdk/knox-android).
> 
> And it was the solution of my problem. All you have to do is add it to your app, create a license and activate it and finally use this method addPackageToBatteryOptimizationWhiteList to unlock the scan when the Samsung device screen is lock.

If you want to scan BLE on a Samsung device when the screen is locked then you need to apply for and install and initialize/whitelist your app in their Knox SDK.

davidgyoung:
@paulpv, I do not think Knox whitelisting is required provided proper scan filters are in place for background scanning.  I have confirmed that background scans are possible on Samsung Galaxy S- devices with Android 8.1-10 provided scan filters are in place.  That said, it may be true that adding a Knox whitelist makes it so you are not required to have the scan filters.  But for most app developers who cannot expect end-users to go to the trouble to whitelist apps,

davidgyoung:
@atotalnoob, I do have a concern with this change:

There are a number of apps out there where scan filters don't work at all.  This happens because the software matching of the library on unfiltered scans is more tolerant than the brute force byte patterns required by hardware scan filters.  Typical reasons for scan filters to fail include:

1. The developer is using a beacon with a non-standard manufacturer code (e.g. iBeacon with a manufacturer code that is not Apple's 0x004c, or AltBeacon with a manufacturer code that is not Radius Networks' 0x0118)
2. The developer is using a custom beacon format with byte patterns that ignore part of the beginning of the advertising packet.

Some apps in the above situation can never get background scan results on Samsung devices, but this is acceptable for them because their use cases primary rely on foreground scanning.  If this change is merged, then those apps will no longer be able to scan in the foreground on Samsung devices.

atotalnoob:
I don't really understand your concern. Up until #855 this was the code path. The reason for #855 had nothing (from what I understand) to do with this issue. So I am a bit confused as to your concern. 

Without this, the library is practically worthless on my samsung tablets. Its causing a lot of issues for me. We are utilizing confidex and custom made beacons, both transmitting as eddystone. I checked, and in 1. we are setting the manufacturer code

davidgyoung:
@atotalnoob, please take a look at #848, which was solved by #855.  If we merge this pull request, I believe this would re-create the problem in #848.

In fact, this PR simply reverts the change in #855, yes?  If not what am I missing?

davidgyoung:
Ok, @atotalnoob, I believe you are saying you think #848 should be reverted because you think the side effects are worse than the problems in #855 that it fixed.

If this is correct, please make your case.   What is the specific technical consequence of #848 that is negative, and why is it worse than the problem described in #855?  Who will it affect?  Try to think about it broadly, not just from your project perspective, but for all users of the library.  

Also, is there not another solution that can solve both problems?

atotalnoob:
@davidgyoung I think the issue is quite clear. Both #914 #933 and this are caused by #848. 

The fix in #848 is more narrow of a use case than mine. He is concerned with specific UID size, whereas my project is scanning with the screen off/on. This, to me, is a much larger use case of your library, rather than a highly specific 128-bit UID issue. 

The negative side of #848 is that empty scan filters scans are started when the tablet screen is ON. If the screen is turned off while it has the empty scan filter it will not pick up any more beacons. This means that *any* user that has a samsung device would not be able to scan with the screen off until the scan restarts. 

I see nothing that indicates this will affect foreground scanning and I would request that you outline why exactly you think that foreground scanning would be affected.

As far as a solution that solves both problems, I would love to hear your ideas, but this is the only solution I have come up with.

davidgyoung:
Yes, @atotalnoob, I understand the problem now.  I do think we need a solution that covers both the  general use case you describe and protects against the failure of s scan filter to match.  

One idea:  If the library listens for screen on/off events, it could force a scan configuration each time.  This may not be so hard to implement as the library already does something similar each time it detects that it has switched from the foreground to the background.  The additional even would fire when it was already in the background, but the screen on/off state changes.

atotalnoob:
@davidgyoung That is similar to what I have been doing as a workaround. Stopping then restarting the scan every time the screen turns off. 

Could you please point me in the direction of the relevant code files and I will create a PR for this?

paulpv:
I am skeptical this would fix much.
I have an app that scans for 3.1 seconds, pauses for 3.1 seconds, repeat until user explicitly "Sign Out" of app.
If restarting a scan after the screen turned off made any difference, then I am already doing that and I wouldn't have the number of users that are complaining that my app cannot scan while their screen is turned off.

By OS scanning rules, scans should not last for more than 30 minutes anyway.
Show me an app that is scanning continuously for anything close to 30 minutes and I'll show you an app that isn't designed well and probably has lots of other problems.
If you are already doing long term scans then you most definitely should be stopping and re-starting them on a regular short duration basis...in which case "Stopping then restarting the scan every time the screen turns off" seems moot to me.

atotalnoob:
I think you should look into the code changes and what `isInteractive()` means. You should also look as to when this change was made (git blame). It was recent and only after that update, did I start experiencing problems specific to this. Knox is an entirely separate issue.  

My company's business problem is to scan continuously. It is done with the users full knowledge and consent. An ad hominem attack is not necessary.

paulpv:
No attack intended.
Just trying to help out; We're both on the same side of Android's BLE mess; I want to find similar solutions to similar problems and help others that I see having similar issues if I can.

I don't know your use case; please don't expect that I do.
A continuous scan is a bad thing.
If you are scanning continuously for more than 30 minutes then your app will, by OS design, go in to Opportunistic Scanning Mode and will not actively scan any more devices...period.
That would be a bad design...period.
Not an attack; a fact that I am just stating without assuming that you do or do not know about it.

I suggest you reconsider the need for a continuous scan, at least for battery usage reasons (but perhaps your use case is plugged in all of the time).
If required, then fine, but do please give it an honest consideration.
3.1 on, 3.1 off, repeat, seems like a reasonable workaround (I do curse Android for implementing the arbitrary "5 scans in 30 seconds" rule).

I am not 100% confident if PendingIntent scans are or are not limited to 30 minutes of active scanning.
I have a hunch that they might not be limited, that only scans using ScanCallback are.
I thought I had tested this before, but cannot confirm from memory.
Perhaps I am wrong though.

For my use case I am also still not convinced that Knox is relevant at all to many of my app's users' claims that their devices cannot be scanned when the Samsung's (and perhaps other manufacturers) screen is locked or off.
I only read about this Samsung hack on the graffiti of the internet within a few days or your posting this PR which came up on the same Google search for the issue; I have not confirmed if this makes any difference or not.
I am just suggesting that **if true** it might help or explain a few things about BLE's background behavior that seems to be prevalent mostly on Samsung devices (which unfortunate for us are the most popular devices).

Working [more or less] directly with large manufacturers [even if only through some "enterprise" intended SDK] to have them not trash our business justified use cases seems a big stretch, but if it does work just might be worth it, especially in niche cases.

atotalnoob:
The library takes care of the 30 minute OS restrictions and protects against most others. Are you using this library? Does that mean the `nrf connect` app scans for only 3.1 on/3.1 off? If so we have been using your app to validate some readings if we don't have a `SDR` setup and this would have clouded our findings...

In addition, I am able to scan for 24+ hours straight and get readings less than `5 seconds` apart, I'd say my solution is working fine, and is not the issue.

A knox license is not very expensive. It seems to be under $100.... https://www.samsungknox.com/en/resellers?solution=knox-workspace

Have nordic buy it and test it out... I'd be interested to hear your results.

paulpv:
Yes, I realized after my posts that these changes are to the `CycledLeScanner*` class.

The `nrf connect` app is Nordic; I am not affiliated with them at all.
I work at pebblebee.com

What is "a `SDR` setup"?

If you are using this library to continuously scan then, as you already mentioned, it already works around the 30 minute limit.
To state the obvious, this library doesn't literally scan for 30 continuous minutes; it just cycles the scan to avoid the limits.
ie: you are good, but any app that **literally** scans for 30 **continuous** minutes is probably bad design and should be revisited.

I can swear that I once proved in a test app that a single PendingIntent scan can run for many hours and continue to receive scan results ignoring the 30 minute scan limit, but I could easily be mistaken.
(I have tried many solutions/test in the past, and they are starting to all blend together in my head...I would need to revisit the app and confirm first hand).

I will revisit this and report back here.
(Sorry if I am hijacking your PR w/ something that may seem unrelated; we're all just trying to help each other out, and a little extra good confirmed proven true info here and there should be a good thing).

I am trying the Knox solution myself and close to finishing and will report back here.
The solution would admittedly suck, especially with the user required to accept a Samsung EULA and add an Admin Profile to the phone, but this whole situation sucks.

I am not actually using this library.
I tried to use this library, but the main reason I am not is that the devices I am scanning are not strictly beacons all of the time, and by the time I created a custom non-beacon scanning pattern I felt like there were better libraries out there for my purposes.
I have watches on all of the Android BLE scanning libraries that I have found (this, RxAndroidBle, Android-Scanner-Compat-Library are the first ones that come to mind), and review any updates to them that I notice.
They all offer a similar scattering of solutions here and there to tons of various Android BLE scanning problems.
None of them seem perfect.
Google/AOSP needs to fix this; no 3rd party will be able to.

I am currently using my own wrapper around https://github.com/NordicSemiconductor/Android-Scanner-Compat-Library, that I have made a few contributions to, and am now considering using SweetBlue if I can prove that it solves more epidemic scanning problems.
For the record I am not yet convinced that the expensive SweetBlue SDK improves things much/any, but when we ask users experiencing problems to try using the SweetBlue Toolbox v3 app, many users' problems [mostly temporarily] go away.
I am convinced that something in the SweetBlue Toolbox v3 app kick-starts many phones to start working.
I am hoping my next code will wrap that behavior.

atotalnoob:
`SDR` is a Software Defined Radio https://en.wikipedia.org/wiki/Software-defined_radio

Yeah, BLE is a mess...

Thought you were from Nordic, since you had contributions to their repos. 

Under the hood, yeah it cycles and does other stuff, but from my codes' perspective its essentially continuous...

davidgyoung:
See proposed alternative solution in #941

paulpv:
FYI, I have confirmed to my satisfaction that there appears to be zero difference when scanning with the screen off/locked with or without Knox.

```kotlin
    fun isSamsung(): Boolean {
        return Build.MANUFACTURER.equals("samsung", ignoreCase = true)
    }

    fun getDevicePolicyManager(context: Context): DevicePolicyManager {
        return context.getSystemService(Context.DEVICE_POLICY_SERVICE) as DevicePolicyManager
    }

    private var _devicePolicyAdmin: ComponentName? = null
    private val devicePolicyAdmin: ComponentName
        get() {
            if (_devicePolicyAdmin == null) {
                _devicePolicyAdmin = ComponentName(this, SampleAdminReceiver::class.java)
            }
            return _devicePolicyAdmin!!
        }

    fun knoxAppPolicy(context: Context): ApplicationPolicy {
        val edm = EnterpriseDeviceManager.getInstance(context)
        return edm.applicationPolicy
    }

    private var _knoxAppIdentity: AppIdentity? = null
    private val knoxAppIdentity: AppIdentity
        get() {
            if (_knoxAppIdentity == null) {
                _knoxAppIdentity = AppIdentity(packageName, null)
            }
            return _knoxAppIdentity!!
        }

    fun knoxAddPackageToBatteryOptimizationWhiteList(activity: Activity, requestCode: Int) {
        if (!isSamsung()) {
            return
        }

        try {
            val packageName = activity.packageName

            if (isKnoxLicenseActivated == null) {
                isKnoxLicenseActivated = false

                deviceAdminRequestActivity = activity
                deviceAdminRequestCode = requestCode

                val licenseManager = KnoxEnterpriseLicenseManager.getInstance(activity)
                licenseManager.activateLicense(SAMSUNG_KNOX_DEV_LICENSE_KEY, packageName)
                return
            }

            if (!isKnoxLicenseActivated!!) {
                return
            }

            val dpm = getDevicePolicyManager(activity)

            isDeviceAdmin = dpm.isAdminActive(devicePolicyAdmin)

            if (false && isDeviceAdmin) {
                dpm.removeActiveAdmin(devicePolicyAdmin)
                isDeviceAdmin = false
            }

            if (!isDeviceAdmin) {
                val intent = Intent(DevicePolicyManager.ACTION_ADD_DEVICE_ADMIN)
                intent.putExtra(DevicePolicyManager.EXTRA_DEVICE_ADMIN, devicePolicyAdmin)
                intent.putExtra(DevicePolicyManager.EXTRA_ADD_EXPLANATION, "TODO: Reason why...")
                activity.startActivityForResult(intent, requestCode)
                return
            }

            val appPolicy = knoxAppPolicy(activity)
            val appIdentity = knoxAppIdentity

            if (true) {
                Log.w(TAG, "knoxAddPackageToBatteryOptimizationWhiteList +appPolicy.addPackageToBatteryOptimizationWhiteList($appIdentity)")
                val result = appPolicy.addPackageToBatteryOptimizationWhiteList(appIdentity)
                Log.w(TAG, "knoxAddPackageToBatteryOptimizationWhiteList -appPolicy.addPackageToBatteryOptimizationWhiteList($appIdentity); result=${appPolicyErrorToString(result)}")
            } else {
                Log.w(TAG, "knoxAddPackageToBatteryOptimizationWhiteList +appPolicy.removePackageFromBatteryOptimizationWhiteList($appIdentity)")
                val result = appPolicy.removePackageFromBatteryOptimizationWhiteList(appIdentity)
                Log.w(TAG, "knoxAddPackageToBatteryOptimizationWhiteList -appPolicy.removePackageFromBatteryOptimizationWhiteList($appIdentity); result=${appPolicyErrorToString(result)}")
            }
        } catch (e: Exception) {
            Log.w(TAG, "knoxAddPackageToBatteryOptimizationWhiteList: Exception: $e")
        }
    }
```

atotalnoob:
That's good to know. 

Have you identified any advantage to integrating Knox?

paulpv:
No direct consumer advantage that I have noticed.
A user would be presented with two nag screens:
  1. Samsun Knox EULA screen
  2. Android Device Admin screen (that theoretically can be limited to a single policy, but following the samples I still always got a screen of a huge list of policies that should scare anyone)
I could never get the Knox AppIdentity signature to work, so I ended up passing `null` and the api call to add the app package to the battery optimization whitelist returned with no error.
Unless passing a `null` signature is invalid, I am satisfied that there is no user benefit.

I am curious if SweetBlue handles Samsung screen off scanning well or has issues too.

atotalnoob:
Have you tried scanning with a non-empty scan filter? That is what this entire PR is about. 

Samsung doesn't allow empty scan filters when the screen is off.

paulpv:
I was speaking from the perspective that Knox would subvert that requirement.

In my current code, currently using https://github.com/NordicSemiconductor/Android-Scanner-Compat-Library with PendingIntent scan, even setting a filter specific for my devices doesn't result in any devices scanned while the screen is off or locked (the latter being even if the screen is reported as `ACTION_SCREEN_ON`)

PR #972, opened 2020-05-22
Add new devices

schauersbergern:
I added new devices, based on this calculations:

https://docs.google.com/spreadsheets/d/185q5AJmWipH23si5IAvMpdy3rGyuJFsCwkqM6EkqFH8

### DIFF
--- a/lib/src/main/resources/model-distance-calculations.json
+++ b/lib/src/main/resources/model-distance-calculations.json
@@ -39,6 +39,72 @@
       "model":"XT1092",
       "manufacturer":"Motorola",
       "default": false
+    },
+    {
+      "coefficient1": 4.882103522,
+      "coefficient2": 6.98084248,
+      "coefficient3": -0.6644164184,
+      "version":"10",
+      "build_number":"QP1A.190711.020",
+      "model":"SM-G980F",
+      "manufacturer":"samsung",
+      "default": false,
+      "name":"Galaxy S20"
+    },
+    {
+      "coefficient1": 4.290357018,
+      "coefficient2": 7.179975601,
+      "coefficient3": -0.4184600771,
+      "version":"9",
+      "build_number":"HUAWEIVOG-L29",
+      "model":"VOG-L29",
+      "manufacturer":"HUAWEI",
+      "default": false,
+      "name":"P30 Pro"
+    },
+    {
+      "coefficient1": 5.166854488,
+      "coefficient2": 6.336069721,
+      "coefficient3": 0.3871575945,
+      "version":"9",
+      "build_number":"HUAWEIMAR-L21A",
+      "model":"MAR-LX1A",
+      "manufacturer":"HUAWEI",
+      "default": false,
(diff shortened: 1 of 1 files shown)


davidgyoung:
Hey, Nice work, @schauersbergern.  

I am a little worried about the -70 dBm RSSI measured at 1 meter on the iPhone 8.  this is quite weak.  A typical 1m signal level for a full power BLE transmitter should be about -60 dBm.  Since it is a logarithmic unit, this suggests the iPhone 8 saw 90% less energy than expected.

One of two things must be true:

1. The iPhone 8 has very poor receiver sensitivity.
2. The beacon used in these tests has a very weak transmitter.

I suspect that (1) is the case because the several of the Android devices you tested did see a signal level of about -60 dBm at 1 meter.  Just to be sure, do you have access to any other iOS devices to take another measurement of your beacon at 1 meter?  (e.g. an iPhone 7, iPhone SE, etc.)

davidgyoung:
@schauersbergern, I did some measurements to compare the receiver sensitivity on various iOS models.  This should allow us to convert measurements by one device to the equivalent by another device.    These measurements include an iPhone SE 1st generation, which has similar hardware to an iPhone 4s-5s that are known to be good reference models.  As you can see from the spreadsheet, an iPhone 6 is a very weak receiver an iPhone 6 is -17 dBm weaker, an iPhone 7 is -2 dBm weaker.  and a Google Pixel 3a is +2 dBm stronger.  If you have any of these phones to get a 1 meter signal measurement off of your test beacon we could make the proper conversions.


https://docs.google.com/spreadsheets/d/1-voNJAeHz78AarZmfds8WiplNhUDzeYrsW0YzKfrExU/edit#gid=0

schauersbergern:
Hello @davidgyoung, thank you for comparing the iOS models!
In our test we used a Samsung tablet with your Locate App with -56 txpower as the beacon advertiser.
I will do a reference measurement with a iPhone 7, but I have 2 questions before:

- Is the Samsung tablet accurate to serve as a beacon?
- Why can't any other phones 1m rssi serve as an reference value?

thank you,
br Nikolaus

davidgyoung:
Any beacon works fine for tests provided that it outputs at least -60dBm of power when measured at 1m.  

Since an iPhone 7 has a slightly insensitive receiver (-2 dB vs iPhone SE 1st Gen) you should measure at least -62 dBm on an iPhone 7 at 1 meter.

The reason we cannot just use any phone to provide the 1m reference value is that all phones have different receiver sensitivity.  We need a common standard that we all use.  The standard was established as an iPhone 4s-5s (insignificant variation between them) early in the beacon era.  The iPhone SE 1st Gen behaves similarly, do it is an effective equivalent.

Can you please get a 1m measurement with both an iPhone 7 and 8 so we can get the delta between them?

schauersbergern:
I think I can organise an iPhone 7 until next week for doing the reference measurement.
Thanks for your patience!

PR #973, opened 2020-05-23
Fix delete and rename scan state file on Windows

DanielBican:
## Error
ScanStateTest.serializationTest is failing on Windows with this error message: "E/ScanState: Error while saving scan status to file: Cannot rename temp file"
It is logged from the save method in ScanState when the temporary file (android-beacon-library-scan-state-temp) is renamed to the normal file (android-beacon-library-scan-state). 
## Cause
The root cause seems to be this error message: "The process cannot access the file because it is being used by another process java.exe" which can be seen when you try to manually rename the file from Windows File System Explorer.
In order to reproduce, place a breakpoint before renameTo is called and then try to apply manual steps.
## Fix
What I think is happening is that the java.exe process is locking the file because the stream resources are not closed so I moved the call to delete and rename the normal file after the resources are closed.
## Gradle Test Report
TestReport230520
## Manual tests
No manual tests.

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
+++ b/lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
@@ -193,17 +193,6 @@ public class ScanState implements Serializable {
                 outputStream = mContext.openFileOutput(TEMP_STATUS_PRESERVATION_FILE_NAME, MODE_PRIVATE);
                 objectOutputStream = new ObjectOutputStream(outputStream);
                 objectOutputStream.writeObject(this);
-                File file = new File(mContext.getFilesDir(), STATUS_PRESERVATION_FILE_NAME);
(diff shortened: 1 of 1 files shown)


davidgyoung:
Thanks for this pull request.  It certainly not makes more sense to delete and rename after closing.  This is clearly triggerefd on Windows due to file system differences.

Hey, @DanielBican, can you please add a try/catch for IOException with logging of an error around the delete/rename?  I would hate to risk triggering intermittent crashes with this change by moving those operations outside such a try/catch.

DanielBican:
Hi David. Just to make sure, it makes sense or not to delete and rename after closing the streams?
Regarding IOException, it is a caught exception and delete or rename are not triggering it. Does it make sense to catch all exceptions?

davidgyoung:
Yes, you are correct that those methods do not throw an IOException.  I forget those are very old Java APIs that use C-like return codes.  We do need to rename and delete, as that is how we copy safely a temporary filename to the final filename.

I think this is OK to merge as-is.

PR #974, opened 2020-05-24
HW filter background time optimization

vladvladau:
### Issue
First beacon detection happens after ~25 minutes after the first run of the application in background mode.

### Cause
It takes ~10 minutes to launch the application after is being killed + 15 minutes to run the first scheduled periodic scan job in order to install the HW filter.

### Fix
After the application is launched in background, do an immediate scan job in order to install immediately the HW filter (which is installed after a scan job life-cycle is completed).
The monitored beacons are detected in 10 minutes (how much it takes to re-launch the application).

The fix is composed from two commits:

1. Initiate background mode before bind in order to avoid 2nd schedule call;
2. Add a _"first time scan flag"_ in order to trigger the immediate scan job if we start in background;

The 2nd commit depends on the 1st one.

### Roboelectric tests

The tests are passed (with fix #973).
image

### Manual tests

Device: Huawei Mate 20 Pro
AndroidOS: 10
Debug without wire (over Wi-Fi)

1. **Before fix**

- BeaconManager started up on pid ...
- This consumer is not bound.
- Not starting beacon scanning service. Using scheduled jobs
- Activating background region monitoring
- Background region monitoring activated for region id1:
- **Applying settings to ScanJob** _//1st run_
- Applying scan job settings with background mode false
- Scheduling immediate ScanJob to run in 50 millis
- Scheduling ScanJob (...) to run every 300000 millis
- updating background flag to true
- updating scan period to 10000, 300000
- **Applying settings to ScanJob** _//2nd run_
- Applying scan job settings with background mode true
- Register calls: global=2 instance=2
- Not scheduling an immediate scan because we are in background mode. Cancelling existing immediate ScanJob.
- Scheduling ScanJob (...) to run every 310000 millis
- consumer count is now: 1
- Waiting for BeaconService connection

2. **After fix**

- BeaconManager started up on pid ...
- updating background flag to true
- updating scan period to 10000, 300000
- This consumer is not bound.
- Activating background region monitoring
- Background region monitoring activated for region id1:
- **Applying settings to ScanJob** _//1st and only run_
- Applying scan job settings with background mode true
- **This is the first time we schedule a job and we are in background, set immediate scan flag to true in order to trigger the HW filter install.** _//new logs_
- Register calls: global=1 instance=1
- We just woke up in the background based on a new scan result or first run of the app. Start scan job immediately.
- Scheduling immediate ScanJob to run in 0 millis
- **First immediate scan job scheduled successful, change the flag to false.** _//new logs_
- Scheduling periodic ScanJob (...) to run every 310000 millis
- **Notify beacon manager that the immediate scan job in background was scheduled successful.** _new logs_
- consumer count is now: 1
- Waiting for BeaconService connection
- Register calls: global=2 instance=2
- Beacon simulator not enabled
- **Started passive beacon scan** _//HW filter is installed after immediate scan job_

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -152,6 +152,7 @@ public class BeaconManager {
     private boolean mRegionStatePersistenceEnabled = true;
     private boolean mBackgroundMode = false;
     private boolean mBackgroundModeUninitialized = true;
+    private boolean mBackgroundScanJobFirstRun = true;
     private boolean mMainProcess = false;
     @Nullable
     private Boolean mScannerInSameProcess = null;
@@ -604,6 +605,9 @@ public class BeaconManager {
     public boolean getBackgroundMode() {
         return mBackgroundMode;
     }
+    public boolean getBackgroundScanJobFirstRun() {
+        return mBackgroundScanJobFirstRun;
+    }
     public long getBackgroundScanPeriod() {
         return backgroundScanPeriod;
     }
@@ -615,6 +619,10 @@ public class BeaconManager {
     }
     public long getForegroundBetweenScanPeriod() {
         return foregroundBetweenScanPeriod;
+    }
+
+    public void setBackgroundScanJobFirstRun(Boolean BackgroundScanJobFirstRun) {
+        mBackgroundScanJobFirstRun = BackgroundScanJobFirstRun;
     }
 
     /**
--- a/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -152,7 +152,6 @@ public class BeaconManager {
     private boolean mRegionStatePersistenceEnabled = true;
     private boolean mBackgroundMode = false;
     private boolean mBackgroundModeUninitialized = true;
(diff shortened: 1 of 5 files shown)


lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
```diff
@@ -617,6 +621,10 @@ public long getForegroundBetweenScanPeriod() {
         return foregroundBetweenScanPeriod;
     }
 
+    public void setBackgroundScanJobFirstRun(Boolean BackgroundScanJobFirstRun) {
```
davidgyoung, line 624: Can we put this flag on ScanJobScheduler instead of BeaconManager?  I would prefer to not put it on BeaconManager since this is a public interface for the library and this flag should not be part of that public interface.  If you move the flag to ScanJobScheduler, then BeaconManager can hold a private reference to the ScanJobScheduler when it uses it, so we make sure we are always using the same instance of ScanJobScheduler.  This way it is logically equivalent to the code you have.

vladvladau: Definitely makes more sense to be like you suggested. I've removed the flag from _BeaconManager_ and _ScanState_ class. Everything is no handled inside _ScanJobScheduler_ class.

davidgyoung:
Thanks @vladvladau!

Issue #990, opened 2020-07-24
startRangingBeaconsInRegion should check range ID


Vandalko:
Long-running app may call "startRangingBeaconsInRegion" a significant number of times.
There's no internal check if such range is already present resulting in unlimited array growth.

I think the correct behavior should be to replace the existing (same ID) region.

davidgyoung:
@Vandalko. I did a quick check and you are absolutely right.  I am surprised that it has been this way for so long!  This does not seem to have caused  any problem reports that I am aware of, but it would still be good to fix.

I'd welcome a pull request if you are up for it.

PR #991, opened 2020-07-24
Replace equals Regions on ranging beacons requests

Vandalko:
Closes https://github.com/AltBeacon/android-beacon-library/issues/990

> Long-running app may call "startRangingBeaconsInRegion" a significant number of times.
There's no internal check if such range is already present resulting in unlimited array growth.


davidgyoung (approved):
Sorry for the late merge.  This is a great change, thank you, @Vandalko

PR #1022, opened 2021-04-05
Remove dependency on androidx.localbroadcastmanager.  Re-enable Maven Central publishing

davidgyoung:
This **removes all code that uses the deprecated class `androidx.localbroadcastmanager.content.LocalBroadcastManager`** which had been removed from the Android 11 SDK.  The  continued use of this class had added the painful need to add an extra dependency on `androidx.localbroadcastmanager` in the `build.gradle` of any project using this library.  That extra dependency is no longer needed.

Two functional areas changed as a result of not having access to `LocalBroadcastManager`:

1. Delivery of ranging and monitoring data from the ScanJob to the client application used to use this construct, but now uses direct method calls using a singleton that brokers the communication.
2. Delivery of BluetoothMedic events to the BluetoothMedic for processing and potential bluetooth power cycling.  Again, this construct has been replaced by direct method calls to the BluetoothMedic singleton.  

There are no known drawbacks to the above refactoring, although if client applications had gone around the public APIs and  relied on the internal (private) library implementation of local broadcasts, this will no longer work.

Separately, this change completely rewrites the build scripts to **once again enable publishing to Maven Central** (since JCenter is going away) and in the process use the latest Gradle and Android SDKs and tools.

**The code from this PR is available on Maven Central as release 2.18-beta.  To use it:**

1. Modify your outermost `build.gradle` to add back in mavenCentral:
 ```
 repositories {
        google()
        jcenter()
        mavenCentral()
    } 
 ```
2. Change your project `build.gradle` files to use the beta library:
```
     implementation 'org.altbeacon:android-beacon-library:2.18-beta'
```

3. Remove the localbroadcastmanager dependency:
```
    // Needed for legacy LocalBroadcastManager used by the library prior to library version 2.18
    //implementation 'androidx.localbroadcastmanager:localbroadcastmanager:1.0.0'
```


davidgyoung:
@EKliptus I believe this is possible and plan to make this change in the next release https://github.com/AltBeacon/android-beacon-library/pull/1023

EKliptus:
@davidgyoung Thanks a lot !

PR #1023, opened 2021-04-28
Reduce min-sdk to 14. 

davidgyoung:
This reduces them minimum Android SDK version for using this library to 14.  

This had been unnecessarily increased to 21 in the 2.18 release.  I could have sworn that I was forced to do this by using a modern Android dependency, but when I went back to find it I could not.


davidgyoung:
This is in 2.19-beta

PR #1035, opened 2021-07-04
Auto bind

davidgyoung:
Two new features:

1. Auto Bind

Adds new methods to start beacon ranging and monitoring that allows forgoing manual calls to `bind(...)` and `unbind(...)` and the use of `BeaconConsumer`.  This simplifies setup.

`beaconManager.startMonitoring(@NonNull Region region)`
`beaconManager.stopMonitoring(@NonNull Region region)`
`beaconManager.startRangingBeacons(@NonNull Region region)`
`beaconManager.stopRangingBeacons(@NonNull Region region)`

The old methods that do require a manual call to bind and a `BeaconConsumer` first still work:

`beaconManager.startMonitoringBeaconsInRegion(@NonNull Region region)`
`beaconManager.stopMonitoringBeaconsInRegion(@NonNull Region region)`
`beaconManager.startRangingBeaconsInRegion(@NonNull Region region)`
`beaconManager.stopRangingBeaconsInRegion(@NonNull Region region)`

2. Adds support for beacon parsers based on 128-bit service UUIDs. Example:

`beaconManager.getBeaconParsers().add(BeaconParser().setBeaconLayout("s:0-15=2F234454-CF6D-4A0F-ADF2-F4911BA9FFA6,m:16-17=A6FF,p:18-18,i:0-15l"))`

In the example above, a beacon with service UUID 2F234454-CF6D-4A0F-ADF2-F4911BA9FFA6 will be matched if the service data has the first two bytes 0xA6 0xFF.  A single identifier will be parsed out of the beacon which will always be the same as the service UUID.  This is useful for treating service advertisements as beacons for some use cases.

This change is available for testing in 2.19-beta5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
+### 2.19-beta5 / 2021-08-01
+
+- Bug fixes to autobind (#1035, David G. Young)
+
+### 2.19-beta4 / 2021-07-10
+
+- Auto start background power saver when using autobind.   Expedite region exits when using intent scan strategy.  (#1035, David G. Young)
+
 ### 2.19-beta3 / 2021-07-04
 
-- Add experimental autobind methods for starting beacon ranging/monitoring without a manual call to bind(). (#10351, David G. Young)
+- Add experimental autobind methods for starting beacon ranging/monitoring without a manual call to bind(). (#1035, David G. Young)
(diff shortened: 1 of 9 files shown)


ajayshekhawat256 (commented):
merge it

PR #1040, opened 2021-07-12
Add immutable flag to pending intents for Android 12

nbadal:
[As required in Android 12.](https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability)
As far as I can tell none of these intents need to be mutable, but let me know if that's wrong!


thebino (approved):
Looks good to me.

CanYumusak:
This PR will lead to the app no longer crashing due to missing flags but does not take into account the new permissions for Android 12. 

These should be taken care of also (in a separate PR I guess?)

davidgyoung:
Thanks, @nbadal.  Have merged this into #1043 with other Android 12 changes.   @CanYumusak please see discussion of permission changes there.

PR #1053, opened 2021-08-17
Fix NullPointerException in IntentScanStrategyCoordinator

PhilipTocsen:
This fixes an Nullpointer Exception that can occur, if `adapter.getBluetoothLeScanner()` returns `null`. The code already checks if `scanner` is not null, but the Type Declaration does not allow that. For the change to work, callback had to be moved inside the if statement.


davidgyoung (approved):
Looks good, thank you, @PhilipTocsen.  Did you see a crash on a specific device or state of a device?

PhilipTocsen:
Good question. I could not trigger the error myself, but our error logging caught a lot cases from Android 8 - 12 and different manufacturers. You can have a look into our sentry, as my name suggests, i'm working at tocsen ;)

davidgyoung:
Ok, please let me know when you have a new release planned and I will roll a new release of this library with the change before then.

Issue #622, opened 2017-11-20
IllegalArgumentException in BeaconManager.startRangingBeaconsInRegion on Android O 8.0.0


dhabensky:
I am getting IllegalArgumentException during call to `startRangingBeaconsInRegion` on Android 8.0.0

Seems like some data is passed incorrectly and thus producing an exception.

Stacktrace:
`android.os.Parcel.readException (Parcel.java:1946)
android.os.Parcel.readException (Parcel.java:1888)
android.app.job.IJobScheduler$Stub$Proxy.schedule (IJobScheduler.java:180)
android.app.JobSchedulerImpl.schedule (JobSchedulerImpl.java:44)
org.altbeacon.beacon.service.ScanJobScheduler.schedule (ScanJobScheduler.java:156)
org.altbeacon.beacon.service.ScanJobScheduler.applySettingsToScheduledJob (ScanJobScheduler.java:85)
org.altbeacon.beacon.service.ScanJobScheduler.applySettingsToScheduledJob (ScanJobScheduler.java:92)
org.altbeacon.beacon.BeaconManager.applyChangesToServices (BeaconManager.java:962)
org.altbeacon.beacon.BeaconManager.startRangingBeaconsInRegion (BeaconManager.java:814)
com.a1s.navimuseum.location.BeaconLocationService.onBeaconServiceConnect (BeaconLocationService.java:142)
org.altbeacon.beacon.BeaconManager.bind (BeaconManager.java:418)
com.a1s.navimuseum.location.BeaconLocationService$1.doInBackground (BeaconLocationService.java:86)
com.a1s.navimuseum.location.BeaconLocationService$1.doInBackground (BeaconLocationService.java:72)
android.os.AsyncTask$2.call (AsyncTask.java:333)
java.util.concurrent.FutureTask.run (FutureTask.java:266)
android.os.AsyncTask$SerialExecutor$1.run (AsyncTask.java:245)
java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1162)
java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:636)
java.lang.Thread.run (Thread.java:764)`

### Expected behavior
Ranging gets started

### Actual behavior
Application crashes

### Steps to reproduce this behavior
bind to `BeaconManager`
call `startRangingBeaconsInRegion(region)`

if it is crucial:
 - `startRangingBeaconsInRegion` is called not in main thread
[...]

PR #1055, opened 2021-08-27
proposal: make JobService's persist option configuable

hotchemi:
related: https://github.com/AltBeacon/android-beacon-library/issues/622

### Summary

There're some cases that we don't need to make the job persisted and `RECEIVE_BOOT_COMPLETED` permission(Not sure how many but at least in our use case it is).
This PR adds `jobPersistedEnabled` option as metadata for `BeaconService` as follows(defaults to true and it keeps backwards compatibility):

```xml
   <service android:name="org.altbeacon.beacon.service.BeaconService"
             tools:node="replace">
        <meta-data android:name="jobPersistedEnabled" android:value="false"/>
    </service>
```

~With this change, the flag would be `false` by default and you need to manually add the option to enable the persistent option.~
~I know breaking backwards compatibility should be avoided and would like to discuss it if you like the idea 🙇~ 

Thx in advance!


hotchemi:
Could you take a look when you have time? 🙏

hotchemi:
Hi~any update? 🙇

davidgyoung:
Hi, @hotchemi.  Apologies, I misunderstood that a key part of this PR is to allow configuration in the manifest that the job is not persisted.  Because I did not understand this, I thought my last suggestion has resolved the issue for you.

Is there a reason that you need this configured in the manifest?  It seems it would be much simpler to have a method to configure this, say:

beaconManager.setJobPersistenceEnabled(false)

Then just set this once in your code if you do not want the default persisted jobs.

Ultimately the library is going to be moving to a new configuration API that is exclusively programmatic (not declared in the manifest), so this might be a better fit.    

Thoughts?

hotchemi:
> Ultimately the library is going to be moving to a new configuration API that is exclusively programmatic (not declared in the manifest), so this might be a better fit.

@davidgyoung Thx for the feedback!

> Is there a reason that you need this configured in the manifest?

This is just to keep the API consistency with [longScanForcingEnabled](https://github.com/AltBeacon/android-beacon-library/pull/529) and I don't have a strong opinion around here🙂 But presumably we should follow after you accomplish migration to the new programmatic API?

We're using the forked version with this change and it's kinda not urgent now, it's also fine closing the PR and wait for your work👍

davidgyoung:
@hotchemi please see the JobSchedulerScanStrategy section [here](https://github.com/AltBeacon/android-beacon-library/issues/1061).  Will this kind of API work for you?

hotchemi:
@davidgyoung Thx, I checked the doc and `scheduledJobScanStrategy.jobPersistenceEnabled = true` looks good to me 👍

hotchemi:
Then let me close the PR. thx!

Issue #1082, opened 2022-04-15
App throws ForegroundServiceStartNotAllowedException when in background (Android 12+)


zahichemaly:
I am testing using the latest BeaconReferenceApplication, with foreground service being enabled.

### Expected behavior
Foreground service should restart after we stop monitoring for a while.

### Actual behavior
App throws `ForegroundServiceStartNotAllowedException` when in the background, and foreground service fails to restart.  

![image](https://user-images.githubusercontent.com/41119320/163577849-5322a578-5ad4-43f7-97d6-63225cbcb4b4.png)

Logs:

> 2022-04-15 16:22:50.628 32675-32675/org.altbeacon.beaconreference I/BeaconManager: BeaconManager started up on pid 32675 named 'org.altbeacon.beaconreference' for application package 'org.altbeacon.beaconreference'.  isMainProcess=true
2022-04-15 16:22:50.635 32675-32675/org.altbeacon.beaconreference D/BeaconParser: Parsing beacon layout: m:2-3=beac,i:4-19,i:20-21,i:22-23,p:24-24,d:25-25
2022-04-15 16:22:50.639 32675-32675/org.altbeacon.beaconreference D/BeaconParser: Parsing beacon layout: m:2-3=0215,i:4-19,i:20-21,i:22-23,p:24-24
2022-04-15 16:22:50.648 32675-32675/org.altbeacon.beaconreference D/CompatibilityChangeReporter: Compat change id reported: 160794467; UID 10555; state: ENABLED
2022-04-15 16:22:50.652 32675-32675/org.altbeacon.beaconreference W/BeaconManager: Disabling ScanJobs on Android 8+ may disable delivery of beacon callbacks in the background unless a foreground service is active.
2022-04-15 16:22:50.656 32675-32675/org.altbeacon.beaconreference I/ScanJob: Using immediateScanJobId from manifest: 208352939
2022-04-15 16:22:50.660 32675-32675/org.altbeacon.beaconreference I/ScanJob: Using periodicScanJobId from manifest: 208352940
2022-04-15 16:22:50.660 32675-32675/org.altbeacon.beaconreference W/BeaconManager: Disabling ScanJobs on Android 8+ may disable delivery of beacon callbacks in the background unless a foreground service is active.
2022-04-15 16:22:50.661 32675-32675/org.altbeacon.beaconreference I/ScanJob: Using immediateScanJobId from manifest: 208352939
[...]

PR #1088, opened 2022-06-03
Handle Foreground Service Start Failures in Android 12

davidgyoung:
Apps using this library on Android 12 can crash if they configure the library's foreground service for scanning as reported in #1082.  This happens because Android 12 restricts when foreground services can be started to when:

* The app is in the foreground
* The app has received BOOT_COMPLETED
* A number of other qualifying events (motion, Firebase messages, Bluetooth connect, etc.) not tracked by the library.

While it is possible to code around the above, the restrictions make it difficult to use the library.  These following changes are designed to make using the library with a foreground service on Android 12 easier.

1. The library will detect if starting its foreground service fails, catching the exception and preventing a crash.
2. If the foreground service can't be started, the library will fall back to using the default ScanJob scanning strategy. This will keep beacon detections working, albeit in a degraded fashion.
3. If the library detects the app has come to the foreground after the above failure has happened, it will start the foreground service at that time.
4. A new public API method on BeaconManager indicates if a failure to start the foreground service has happened:  foregroundServiceStartFailed()
5. A new public API method on BeaconManager allows the app to retry starting the foreground service (useful for the case that a qualifying event other than the automatically handled foreground change is  known to have happened): retryForegroundServiceScanning()

For this change I performed the following tests with the Kotlin reference app:

1. Modifying app to use a forreground service but long background between scan periods
2. Launching app with no beacons around
3. Killing foreground service from adb while app is in a between scan cycle.
4. Turning on a beacon to re-launch the app in the background. I see this:

```
2022-06-03 11:47:45.897 5802-5802/org.altbeacon.beaconreference I/BeaconManager: Attempting to starting foreground beacon scanning service.
2022-06-03 11:47:45.905 5802-5802/org.altbeacon.beaconreference W/BeaconManager: Foreground service blocked by ServiceStartNotAllowedException.  Falling back to job scheduler
```

The above confirms fallback to ScanJobs works.

Now test failover when the app. comes to the foreground:

5. Tapp app launcher to bring to foreground

```
06-03 11:52:38.440  5802  5802 I BeaconManager: unbinding all consumers for strategy failover
06-03 11:52:38.440  5802  5802 D BeaconManager: Unbinding
06-03 11:52:38.440  5802  5802 D BeaconManager: Not unbinding from scanning service as we are using scan jobs.
06-03 11:52:38.440  5802  5802 D BeaconManager: Before unbind, consumer count is 1
06-03 11:52:38.440  5802  5802 D BeaconManager: After unbind, consumer count is 0
06-03 11:52:38.440  5802  5802 I BeaconManager: Cancelling scheduled jobs after unbind of last consumer.
06-03 11:52:38.440  5802  5802 I ScanJob : Using immediateScanJobId from manifest: 208352939
06-03 11:52:38.442  5802  5802 I ScanJob : Using periodicScanJobId from manifest: 208352940
06-03 11:52:38.442  5802  5802 I BeaconManager: binding all consumers for strategy failover
06-03 11:52:38.443  5802  5802 D BeaconManager: Need to rebind for switch to foreground service
06-03 11:52:38.443  5802  5802 D BeaconManager: Binding to service
06-03 11:52:38.443  5802  5802 I BeaconManager: Attempting to starting foreground beacon scanning service.
06-03 11:52:38.444  5802  5802 I BeaconManager: successfully started foreground beacon scanning service.
06-03 11:52:38.446  5802  5802 D BeaconManager: consumer count is now: 1
06-03 11:52:38.446  5802  5802 I BeaconManager: Done with failover
```

Verify foreground service is started and beacons are being scanned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 2.19.5-beta1 / 2022-04-14
+
+- Add checking to see if ThreadPoolExecutor is stalled and try to recover. (TBD, David G. Young)
+
 ### 2.19.4 / 2022-03-10
 
 - Add ApiTrackingLogger (#1078, David G. Young)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### 2.19.5-beta1 / 2022-04-14
 
-- Add checking to see if ThreadPoolExecutor is stalled and try to recover. (TBD, David G. Young)
+- Add checking to see if ThreadPoolExecutor is stalled and try to recover. (#1081, David G. Young)
 
 ### 2.19.4 / 2022-03-10
 
--- a/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -1057,9 +1057,7 @@ public class BeaconManager {
     }
 
     /**
-     * Call this method if you are running the scanner service in a different process in order to
-     * synchronize any configuration settings, including BeaconParsers to the scanner
-     * @see #isScannerInDifferentProcess()
+     * Call this method in order to apply your settings changes to the already running scanning process
      */
     public void applySettings() {
         LogManager.d(TAG, "API applySettings");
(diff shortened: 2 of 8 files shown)


davidgyoung:
This has been moved to #1090 with a cleaner set  of changes.

PR #1091, opened 2022-06-05
Fix bugs with changing BeaconParsers for running scan service

davidgyoung:
This fixes a number of bugs with the way the library applies settings to a running scan service, particularly BeaconParsers.   Prior to this change, apps using a Foreground Service that attempted to add or remove a new BeaconParser after scanning had already started, failed to do so, and race conditions could cause the BeaconParser list to be stuck with an empty list blocking detections.

After this change, configuration changes are automatically applied to a running service and the public API method. `beaconManager.applySettings()` can once again be used to force apply changes to running services. (Calling that method directly is not necessary as it is automatically called if needed when you make config changes.)

Since Android 8, the use of a Scanning Service usually only applies to a Foreground Service, as long-running scanning services are forbidden on Android 8+.  The Job Scheduler is used by default on Android 8+ and the problems described in this PR do not apply to using the Job Scheduler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 2.19.5-beta1 / 2022-04-14
+
+- Add checking to see if ThreadPoolExecutor is stalled and try to recover. (TBD, David G. Young)
+
 ### 2.19.4 / 2022-03-10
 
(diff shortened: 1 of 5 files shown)


davidgyoung:
This is in 2.19.5-beta5

PR #1098, opened 2022-07-26
Expose raw bytes of beacon and accept AD type 0x07

davidgyoung:
* Exposes raw bytes of beacon, useful for troubleshooting purposes
* Accept AD type 7 for 128 bit service UUID beacon detections


davidgyoung:
Published as 2.19.5-beta8

Issue #1109, opened 2022-10-27
BluetoothMedic's power cycle workaround no longer works on Android 13


mannodermaus:
First off, this is not a bug report: I'm not positive that there's anything that can be done by `BluetoothMedic` on devices running Android 13 or above. Consider this more of an FYI with the potential to improve the library's API to account for shortcomings of Android!

---

Basically, every once in a while my Pixel 6 Pro runs into a state wherein any BLE scan yields a `SCAN_FAILED_APPLICATION_REGISTRATION_FAILED` error. I have `BluetoothMedic`'s automatic power-cycles enabled on this project and the logs confirmed that this was attempted correctly, but the issue wasn't resolved for some reason. I need to turn Bluetooth off and back on by hand, and then all works well again.

Checking the Android documentation, I learnt that both `BluetoothAdapter.disable()` and `BluetoothAdapter.enable()` were deprecated without replacement in API 33. Neither method does anything when called on a device running it ([source](https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#disable())). I have yet to check if this is related to the project's `targetSdk` version, but I suspect that it's a "feature" of Android 13 in general.

In essence, `BluetoothMedic.enablePowerCycleOnFailures()` can no longer be relied upon for Android 13 and beyond. I was thinking about a potential replacement API, but all I can think of is to present some UI to the user and inform them about the broken Bluetooth state (`setNotificationsEnabled()` works, but I cannot localize the text of the notifications as they are hardcoded). _I'm wondering if an optional listener interface for broken Bluetooth state should be added to `BluetoothMedic`._ An application could hook into the checks and react to bad states. Maybe it makes sense to deprecate the current variant of `enablePowerCycleOnFailures()` and redirect to a version that requires this listener, allowing for some sort of fallback on newer devices.

### Expected behavior
The best possible recourse to fix bad BLE states on Android 13 is provided via `BluetoothMedic`.

### Actual behavior
On the surface, it looks like BLE power-cycling is done correctly. However, Android 13 silently fails any call to programmatically change the state of a `BluetoothAdapter`.

### Steps to reproduce this behavior
I don't have any reliable steps to get a device into the bad state related to `SCAN_FAILED_APPLICATION_REGISTRATION_FAILED`. But once done, `BluetoothMedic` is unable to get the device out of that state if that device runs on API 33.

### Mobile device model and OS version
Pixel 6 Pro, Android 13

### Android Beacon Library version
2.19.4

[...]

PR #1121, opened 2023-01-06
Prevent crashes in BluetoothMedic from missing Android 12 permissions

davidgyoung:
- also removed unused broadcast notification code

This addresses the issue reported here: https://stackoverflow.com/questions/74709808/altbeacon-bluetoothmedic-crashes-with-android-version-12-13

This also deprecates the BluetoothMedic’s power cycling method in favor of a renamed legacy method that makes it clear that power cycling is no longer possible on Android 13 per #1109

### DIFF
--- a/lib/src/main/java/org/altbeacon/bluetooth/BluetoothMedic.java
+++ b/lib/src/main/java/org/altbeacon/bluetooth/BluetoothMedic.java
@@ -158,15 +158,34 @@ public class BluetoothMedic {
         }
     }
 
+
+
     /**
      * If set to true, bluetooth will be power cycled on any tests run that determine bluetooth is
-     * in a bad state.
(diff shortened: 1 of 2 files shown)


lib/src/main/java/org/altbeacon/bluetooth/BluetoothMedic.java
```diff
@@ -193,20 +158,39 @@ private void initializeWithContext(Context context) {
...
     @RequiresApi(21)
+    @Deprecated
     public void enablePowerCycleOnFailures(Context context) {
+        legacyEnablePowerCycleOnFailures(context);
+    }
+
+
+    /**
+     * If set to true, bluetooth will be power cycled on any tests run that determine bluetooth is
+     * in a bad state.  This only works on Anroid 4.3-12.x devices, as the ability to power cycle
```
tstump-phunware, line 181: nit: typo in the word Android

Issue #1129, opened 2023-02-15
Crash on ScanState.restore


03July:
I have many crashes on :
Fatal Exception: java.lang.IllegalStateException
unread block data
org.altbeacon.beacon.service.ScanState.restore (ScanState.java:145)

This happen after the application update, it seems that the file "android-beacon-library-scan-state" is corrupted or incompatible with the update.

The exception should be caught and should be handled as if the serialized ScanState does not exist rather letting the app crash.

Beacon Library version: 2.19.5
This problem occurs on many manufacturer devices: Samsung, OPPO, Xiaomi, Huawei, Sony....
OS versions: 8,9,10,11,12 & 13

The full stack is 
java.io.ObjectInputStream$BlockDataInputStream.setBlockDataMode (ObjectInputStream.java:2602)
java.io.ObjectInputStream.readObject0 (ObjectInputStream.java:1472)
java.io.ObjectInputStream.defaultReadFields (ObjectInputStream.java:2142)
java.io.ObjectInputStream.readSerialData (ObjectInputStream.java:2066)
java.io.ObjectInputStream.readOrdinaryObject (ObjectInputStream.java:1927)
java.io.ObjectInputStream.readObject0 (ObjectInputStream.java:1440)
java.io.ObjectInputStream.readObject (ObjectInputStream.java:428)
org.altbeacon.beacon.service.ScanState.restore (ScanState.java:145)
org.altbeacon.beacon.service.ScanJobScheduler.applySettingsToScheduledJob (ScanJobScheduler.java:102)
org.altbeacon.beacon.BeaconManager.applyChangesToServices (BeaconManager.java:1368)
org.altbeacon.beacon.BeaconManager.startMonitoringBeaconsInRegion (BeaconManager.java:1209)
org.altbeacon.beacon.startup.RegionBootstrap$InternalBeaconConsumer.onBeaconServiceConnect (RegionBootstrap.java:218)
org.altbeacon.beacon.BeaconManager.bindInternal (BeaconManager.java:462)
org.altbeacon.beacon.BeaconManager.bind (BeaconManager.java:427)
org.altbeacon.beacon.startup.RegionBootstrap.<init> (RegionBootstrap.java:140)
com.k.basemanager.BaseManager$1.onSuccess (BaseManager.java:421)
com.k.basemanager.BaseManager$1.onSuccess (BaseManager.java:2)
com.google.common.util.concurrent.Futures$CallbackListener.run (Futures.java:1076)
[...]

PR #1130, opened 2023-03-03
Partial fix for ScanState.restore IllegalStateException

paolorotolo:
This PR adds `IllegalStateException` to caught exceptions during state restoring and considers them as deserialization errors.

This doesn't fix all the issues of #1129 like excessive file growth, but it's a quick fix to avoid crashes in existing apps.


paolorotolo:
Fixed in #1131

PR #1131, opened 2023-03-03
WIP fixing crash on scan state restore.  

davidgyoung:
This is a work in progress on fixing problems in #1129.

Changes compile and unit tests still pass, but end-to-end testing has not started.

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
+++ b/lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
@@ -38,11 +38,17 @@ public class ScanState implements Serializable {
     private static final String STATUS_PRESERVATION_FILE_NAME = "android-beacon-library-scan-state";
     private static final String TEMP_STATUS_PRESERVATION_FILE_NAME = "android-beacon-library-scan-state-temp";
     public static int MIN_SCAN_JOB_INTERVAL_MILLIS = 300000; //  5 minutes
+    private static long SANITY_FILE_SIZE_LIMIT = 100000; // ~100k
+    private static long SANITY_REGION_LIMIT = 1000;
+    private static long SANITY_PARSER_LIMIT = 1000;
 
     private Map<Region, RangeState> mRangedRegionState = new HashMap<Region, RangeState>();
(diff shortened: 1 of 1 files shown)


lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
```diff
@@ -140,13 +146,26 @@ public static ScanState restore(Context context) {
...
-                inputStream = context.openFileInput(STATUS_PRESERVATION_FILE_NAME);
-                objectInputStream = new ObjectInputStream(inputStream);
-                scanState = (ScanState) objectInputStream.readObject();
-                scanState.mContext = context;
-            } catch (FileNotFoundException fnfe) {
+                File file = context.getFileStreamPath(STATUS_PRESERVATION_FILE_NAME);
+                if (file.length() > SANITY_FILE_SIZE_LIMIT) {
+                    // make sure file size is reasonable.  If over 100k, do not restore
+                    // See issue #1129
+                    LogManager.e(TAG, "Refusing to restore file of size "+file.length());
```
paolorotolo, line 153: Could we maybe delete `STATUS_PRESERVATION_FILE_NAME` if it is too big or corrupted?

paolorotolo:
So I did some tests with multiple files and it correctly marks them as corrupted or skips them for excessive size. It seems it's working fine.

lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
davidgyoung: I think the code does something similar as is because as soon as the file is persisted (typically after app processing a few milliseconds later) the scan state is persisted with the newly emptier state, overwriting the problematic file.

davidgyoung:
This is being published as 2.19.6-beta3

liedQM:
@davidgyoung any update on this PR? I guess the verifying should have taken place now? Can you please also have a look, why `restore` is called from the MainThread causing ANR? The restore is accessing a file so IO operation and should *never* be called from a Main Thread.

liedQM:
@davidgyoung do you have an update? We're still seeing the issue in our stack.

PR #1135, opened 2023-03-14
Add lastPacketRawBytes when using copyBeaconFields

paolorotolo:
When using `Beacon.Builder().copyBeaconFields()`, beacon's raw data is not copied in the new object. This PR fixes that and updates copyBeaconFields test cases.


davidgyoung (approved):
Looks great, @paolorotolo.  Thanks for the contribution.

PR #1138, opened 2023-04-05
refactor

harmit011:
I have performed some refactorings on few files. They are listed below:
        modified:   lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
        modified:   lib/src/main/java/org/altbeacon/beacon/BeaconParser.java
        new file:   lib/src/main/java/org/altbeacon/beacon/NotifierManager.java
        modified:   lib/src/main/java/org/altbeacon/beacon/service/BeaconService.java
        modified:   lib/src/main/java/org/altbeacon/beacon/service/ScanJob.java
        modified:   lib/src/main/java/org/altbeacon/beacon/service/ScanState.java
        modified:   lib/src/main/java/org/altbeacon/beacon/startup/RegionBootstrap.java
        
I'm interested in contributing more as per my time availability

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
+++ b/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java
@@ -472,25 +472,7 @@ public class BeaconManager {
                         }
                         else {
                             LogManager.i(TAG, "Attempting to starting foreground beacon scanning service.");
-                            try {
-                                mContext.startForegroundService(intent);
-                                if (mScheduledScanJobsEnabledByFallback) {
-                                    LogManager.i(TAG, "Successfully switched to foreground service from fallback");
-                                    mScheduledScanJobsEnabledByFallback = false;
-                                    ScanJobScheduler.getInstance().cancelSchedule(mContext);
-                                }
-                                else {
-                                    LogManager.i(TAG, "successfully started foreground beacon scanning service.");
-                                }
-                            }
-                            catch (ServiceStartNotAllowedException e) {
-                                // This happens on Android 12+ if you try to start a service from the background without a
-                                // qualifying event
-                                LogManager.w(TAG, "Foreground service blocked by ServiceStartNotAllowedException.  Falling back to job scheduler");
-                                mScheduledScanJobsEnabledByFallback = true;
-                                syncSettingsToService();
-                                return;
-                            }
+                            if (attemptStartForeground(intent)) return;
                         }
 
                     }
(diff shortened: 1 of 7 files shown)


harmit011:
Extract Method
Location: 
File name : BeaconManager.java
Class: BeaconManager
Method: bindInternal

Explanation: Extracted a method “attemptStartForeground” which takes out complex functionality of try catch block into a separate method. Reduces complexity of method

Decompose Conditional
Location: 
File name : BeaconParser.java
Class: BeaconParser
Method: fromScanData

Explanation: Decomposed complex if else block into 3 methods and optimized code readability 

Introduce Explaining Variable
Location:
File name: BeaconManager.java
Class: BeaconManager
Method: unbindInternal

Explanation: Introduced an explaining variable “scanEnabled” to remove complex conditional code smell

Move method/field
Location:
File name: Scanjob.java -> ScanState.java
Class: Scanjob, ScanState
Method: initialzeScanHelper -> getScanHelperInitialized

Explanation: Moved sections of method initializeScanHelper to a new method getScanHelperInitialized method, removes feature envy smell

Extract class
Location:
File name: BeaconManager.java
Class: BeaconManager
Method: setRangeNotifier, addRangeNotifier, removeRangeNotifier, removeAllRangeNotifiers, setMonitorNotifier, addMonitorNotifier, removeMonitorNotifier, removeAllMonitorNotifiers

New Class: NotifierManager.java

Explanation: A range of notifier methods were present in BeaconManager class these were breaking SRP principle and were making class in-cohesive. Created a new class NotifierManager to move this functionality to a new class.
Reduces the size of class

Replace conditional with polymorphism

Location:
File name: BeaconService.java
Class: IncomingHandler
Method: handleMessage

Explanation: Conditional switch statement with complex calls have been replaced with polymorphism. Created an abstract class and called abstract method of class to get relevant objects

davidgyoung:
Thanks for the PR, @harmit011. Can you please summarize any testing you have done to ensure these refatorings do not alter functionality?

harmit011:
I ran the existing test cases with ".gradlew test" and all test cases passed, even the build succeeded. Since, I didn't introduce a new functionality, passing existing test cases felt sufficient

davidgyoung:
@harmit011, unfortunately this library doesn't have anywhere near 100% code coverage so confirming that unit tests pass does not confirm that a refactoring does not cause a change.  You must test any refactoring not covered by existing tests by building and running the Reference App with any library changes and confirming beacon detection features still work.

There are two main reasons for the library's low test coverage (1) Bluetooth APIs interact with the hardware layer and are very difficult to test effectively without a sophisticated mocking layer that emulates the Bluetooth stack, and building such a mocking layer is probably more effort than the library itself.  (2) Android APIs, particularly Service APIs, are not easily covered in unit tests and Robolectric is used to at least allow calling them without throwing exceptions.

Such interactive testing of library changes requires a good working knowledge of the library. You really need to have built and tested at least one or two Bluetooth beacon apps with the library to be able to do this.  Have you done this?

Right off, I do see one change in this PR that appears to break basic library functionality.  As far as I can see, the new  NotifierManager is not referenced from BeaconManager at all, so as of the change in the PR, there is now no way to set the notifiers.  Further, moving methods to a new class from BeaconManager changes the API and breaks backward compatibility, something we just can't do.  On top of that, it seems that the Set<RangeNotifier> and Set<MonitorNotifier> instances have simply been copied from BeaconManager to NotifierManager and the accessors moved -- this won't work.  It effectively makes it so the copies on BeaconManager are impossible to modify, and modifying the ones on NotifierManager does nothing.

The above problems are not caught by unit tests because there is 0% coverage for the methods that modify the notifiers.

I think this is probably too ambitious of a set of changes for a first PR.  Before submitting a PR you really should use the library to build a beacon app and learn how to test it end-to-end.  Once you have done that, you can try submitting a smaller PR, but if refactoring code, I would focus on one functional part of one class at a time to seep the changes small, and you need to either (a) ensure that existing test coverage exists for the parts of the class changed (b) add that test coverage yourself or (c) do an interactive end-to-end test with actual Bluetooth beacons that ensures nothing is broken by the refactoring, and include a description of these interactive tests with the PR.

I will close this PR for now, but I would welcome a new one if you can make one consistent with what is described above.

Issue #1115, opened 2022-12-19
beacon distance value showing negative distance on Motorola phones (moto edge 20 pro & moto g5+)


gptshubham595:
### Expected behavior
    Distance observed via beacon should be positive

### Actual behavior
       Distance observed via beacon is negative at any distance even if there's a little shake

### Steps to reproduce this behavior
      1. Install altbeacon reference app on mototrola phone and observe the distance shown
      2. shake the phone

### Mobile device model and OS version
     Tested on moto edge 20 pro & moto g5+
     OS Version: 12
     
### Android Beacon Library version
   altbeacon_version = "2.19.4"

davidgyoung:
Thanks for this report.   Can you please provide a few samples of these two values when you see the negative distance?

1. beacon.runningAverageRssi
2. beacon.distance

gptshubham595:
Tested on Moto G5 S+
<google-sheets-html-origin>

13:37:28 | Distance: | 1.358780418
-- | -- | --
13:37:28 | Running | -87
13:37:29 | Ranged: | 1
13:37:29 | Distance: | 1.042936817
13:37:29 | Running | -85
13:37:30 | Ranged: | 0
13:37:30 | Ranged: | 0
13:37:31 | Ranged: | 0
13:37:32 | Ranged: | 1
13:37:32 | Distance: | 0.7765247283
13:37:32 | Running | -83
13:37:32 | Ranged: | 1
13:37:32 | Distance: | -0.1508527346
13:37:32 | Running | -70
13:37:33 | Ranged: | 1
13:37:33 | Distance: | -0.1133826044
13:37:33 | Running | -71
13:37:34 | Ranged: | 1
13:37:34 | Distance: | -0.2150709093
13:37:34 | Running | -68
13:37:35 | Ranged: | 0
13:37:35 | Ranged: | 1
13:37:35 | Distance: | 0.9506274682
13:37:35 | Running | -65.66666667
13:37:36 | Ranged: | 0
13:37:37 | Ranged: | 1
13:37:37 | Distance: | -0.2590933289
13:37:37 | Running | -66.33333333
13:37:38 | Ranged: | 0
13:37:39 | Ranged: | 1
13:37:39 | Distance: | -0.2052836873
13:37:39 | Running | -68.33333333
13:37:40 | Ranged: | 1
13:37:40 | Distance: | -0.242427232
13:37:40 | Running | -67
13:37:40 | Ranged: | 0
[...]

PR #1151, opened 2023-07-06
Improved model-distance-calculations.json

gptshubham595:


gptshubham595:
Added a python script to get the coefficients value, and a sample model.json value

davidgyoung:
Thanks for this update, @gptshubham595 .  That python script will be useful.  Do you think you can:

1. Move the python script to a new folder: lib/scripts so it is outside the Android source tree
2. Write up instructions in a README.md in that same folder, using the info in this page as a starting point: https://altbeacon.github.io/android-beacon-library/distance-calculations2.html

Having instructions will make it easier for other folks to set up the dependencies for the python script, then use it to get new coefficients to be added to the library's config file.  Thank you!

gptshubham595:
I've added all the changes!  #1115 
@davidgyoung

PR #1153, opened 2023-08-14
Backward compatibility - add 'fromScanData' to AltBeaconParser and BeaconParser

drewster99:
Add back original/older method signatures which don't require the `timestampMs` field, to maintain backward compatibility with older closed-source clients which depend on android-beacon-library.

### DIFF
--- a/lib/src/main/java/org/altbeacon/beacon/AltBeaconParser.java
+++ b/lib/src/main/java/org/altbeacon/beacon/AltBeaconParser.java
@@ -57,6 +57,22 @@ public class AltBeaconParser extends BeaconParser {
      * Construct an AltBeacon from a Bluetooth LE packet collected by Android's Bluetooth APIs,
      * including the raw Bluetooth device info
      *
+     * timestampMs excluded for backward compatibility with older api consumers.
+     *
+     * @param scanData The actual packet bytes
+     * @param rssi The measured signal strength of the packet
+     * @param device The Bluetooth device that was detected
+     * @return An instance of an <code>Beacon</code>
+     */
+    @Override
+    public Beacon fromScanData(byte[] scanData, int rssi, BluetoothDevice device) {
+        return fromScanData(scanData, rssi, device, 0, new AltBeacon());
+    }
+
+    /**
+     * Construct an AltBeacon from a Bluetooth LE packet collected by Android's Bluetooth APIs,
+     * including the raw Bluetooth device info
+     *
      * @param scanData The actual packet bytes
      * @param rssi The measured signal strength of the packet
      * @param device The Bluetooth device that was detected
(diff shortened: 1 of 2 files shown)


davidgyoung:
Thanks for the PR, @drewster99. Can you please explain the problem this solves?  How specifically are older clients broken with the library now?

drewster99:
Hi @davidgyoung -- thanks for the reply!

 I've got a closed source client that's built against -- I think it's built against 2.9.1.  
The `fromScanData` method signature was changed a while ago to add an additional parameter -- `timestampMs`.  The closed source client we've got thus can't find `fromScanData` if we use the current altbeacon library.  

The old 2.9.1 library failed because the overall app is now being built targeting API level 33, so `FLAG_MUTABLE` or `FLAG_IMMUTABLE` is now required on PendingIntent.  Because we can't alter the source of the thing that's interfacing with android-beacon-library, we either need to modify 2.9.1 to add `FLAG_IMMUTABLE` (or mutable), or modify the latest and greatest to add back the now-missing `fromScanData` method signature.

For now we've used our forked copy of the latest with the `fromScanData` change.  This seems to work, except that we now get very long delays (maybe 15 seconds?) to find our beacons -- an issue we didn't have previously.  Any thoughts on that?  Did something change fundamentally about the way to use the library along the way?  

If we use the 2.9.1 library adding the `FLAG_MUTABLE` -- for one it's hard to build, but my concern is that we'll run into some other unexpected issue soon enough. 

(I know I know... we need to get that 'middle' library out of the picture, but one thing at a time!)

THANK YOU!

davidgyoung:
Thanks for the explanation, @drewster99.  This makes sense.

While admittedly not well documented, calling the BeaconParser fromScanData method directly is not intended to be a public API, which is why backward compatibility was broken with the timestamp change.  That said, so long as you are willing to take the risk of this changing in the future, you are welcome to call it directly.

I am OK with merging this in, but I have two suggestions:

1. What do you think about defaulting the timestamp to the current time instead of 0?  It seems this would cause fewer unexpected results.
2. I'd suggest marking the methods you added with @deprecated annotation so it is clear that folks should be cautious about using them.  (Although given that this is not intended to be a public API anyway, the value of this may be minimal.)

With respect to the 15 second delay you mention, I am not sure I understand how this comes up -- are you even using the library to initiate BLE scans?  Given that you are calling `fromScanData` directly, I had assumed you were using your own scan and then just using the library for beacon parsing.

drewster99:
Hi @davidgyoung -- thanks for the feedback and suggestions.

I'm in full agreement with both your suggestions (1- defaulting to current time instead of 0, and 2- marking the added methods as @deprecated).  Would you like me to make the changes, or you?

We're using the library two different ways for different purposes – in one way, we _are_ using it to initiate scans, but in another we are not.  I wasn't able to reproduce the long delay again, so we can chalk that one up to cosmic rays at this point. 😀 

Thank you again!

davidgyoung:
Thanks, Andrew.  Yes, a pull request would be most welcome!

drewster99:
@davidgyoung -- existing PR updated, adding the default value for timestampMs and the @deprecated.  Thanks!

PR #1155, opened 2023-09-02
Update foreground service support for Android 14 restrictions

davidgyoung:
This updates the library to support the upcoming Android 14 release, which imposes new restrictions on starting a foreground service.

This adds the necessary AndroidManifest.xml declarations for foreground services on Android 14, and also checks that `Manifest.permission.ACCESS_FINE_LOCATION` has been granted by the user in the call to `beaconManager.enableForegroundServiceScanning(...)`. If permission is not obtained when this call is made, the method now throws a `SecurityException` to let the user know this is a requirement.  

I have tested this on a Pixel 7 running Android 14 Beta 5 using a special branch of the Kotlin reference app and confirmed that the SecurityException crashes the app as described above before permission is granted, and it works to detect beacons with the Foreground service if the call is made after permission is granted.

It is unfortunate that this solution crashes the app rather than gracefully recover.  But at least the SecurityException happens during a library call, with a helpful message about what the developer needs to do, vs. during the OS startup of the service with a less helpful message.

A more graceful recovery would mean one of two things:

1. Don't crash the app, but leave beacon scanning broken.  (I like this solution even less, as it will hide the problem for many developers.)
2. Avoid a crash by not starting the foreground service if permission is not granted, then listen for the permission being granted, and if it is granted, start the foreground service at that time.

The problem with option 2 is that it is very complicated -- it is not easy to listen for this permission being granted.  And then the library gets in the business of managing permissions, something it does not do now.  Any alternative thoughts appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+
+### 2.20-beta1 / 2023-09-02
+
+- Updates to support Android 14 foreground services. (#1155, David G. Young)
+
 ### 2.19.6 / 2023-07-02
 
 - Fix scans being stuck on when Bluetooth turned off in quick settings on
--- a/README.md
+++ b/README.md
@@ -50,8 +50,7 @@ replacing `${altbeacon.version}` with the version you wish to use.
 
 ## How to build this Library
 
-This project uses an AndroidStudio/gradle build system and is known working with Android Studio
-4.1.3 and Gradle 6.5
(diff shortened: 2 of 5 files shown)


davidgyoung:
[2.20-beta1 release has this change](https://mvnrepository.com/artifact/org.altbeacon/android-beacon-library/2.20-beta1) and the kotlin reference app has been updated to use the beta version of the library and change the permissions flow for foreground services [here](https://github.com/davidgyoung/android-beacon-library-reference-kotlin/tree/android-14)

Issue #1061, opened 2021-11-04
Improved Settings and Region configuration APIs


davidgyoung:
# The Problem

Configuring the library has evolved over time with a mixed bag of styles of settings including:

* settings in manifest declarations
* settings on the BeaconManager (both static and instance)
* settings on Beacon (static)
* BeaconParsers

In addition, changing settings at runtime (after scanning has started) can sometimes cause problems and behavior is either poorly documented, non-deterministic, or at worst cause crashes.  Even in the best of cases, changes to not necessarily take effect right away, and it is unclear to users what does take effect right away and what does not.

## Constraints

This proposal for improvement in the library API is designed to stay within the following constraints:

1. Make it easier for library users to configure library settings
2. Retain existing settings API (temporarily for 2.x library versions with API marked as deprecated) for backward compatibility.  
3. Only support the new settings API in 3.0+ library releases

## Settings API Proposal

The Settings Class below is used to set up library scanning settings.  These are typically set before starting ranging/monitoring.   If you want to change them after ranging and monitoring are already started, the settings change will stop and restart ranging and monitoring.

The class design below is designed to use a Kotlin data class  with a Builder class designed for use by Java apps

Kotlin:
```
        val settings = Settings(
            debug = true,
            distanceModelUpdateUrl = "https://s3.amazonaws.com/android-beacon-library/android-distance.json",
            scanStrategy = Settings.ForegroundServiceScanStrategy(
                Notification.Builder(
                    context,
                    "BeaconReferenceApp"
                ).build(), 1
            ),
            longScanForcingEnabled = true,
            scanPeriods = Settings.ScanPeriods(
                foregroundScanPeriodMillis = 1100,
                foregroundBetweenScanPeriodMillis = 0,
                backgroundScanPeriodMillis = 1100,
                backgroundBetweenScanPeriodMillis = 0))
        beaconManager.adjustSettings(settings)
```

Java:
```
        Settings settings = new Settings.Builder()
                .setDebug(true)
                .setDistanceModelUpdateUrl("https://s3.amazonaws.com/android-beacon-library/android-distance.json")
                .setScanPeriods(new Settings.ScanPeriods(1100, 0, 10000, 0))
                .setScanStrategy(new Settings.ForegroundServiceScanStrategy(
                        new Notification.Builder(context, "BeaconReferenceApp").build(),1)
                )
                .setLongScanForcingEnabled(true)
                .build();
        beaconManager.adjustSettings(settings); 
```

[...]

PR #1177, opened 2024-01-14
New Configuration API for 3.0

davidgyoung:
This working branch implements the new Settings and BeaconRegion configuration APIs that are planned for 

The new APIs are expected to live alongside the old APIS for future 2.x releases of the library
Only the new APIs will exist in the 3.0+ versions of the library

See #1061 for design details


davidgyoung:
This is in 2.21.0-beta2 and higher

Issue #1194, opened 2024-05-28
ScanJob restarts scanning every time on navigation


VolodaUa:
### Expected behavior
The scan job should not restart every time once you have next/back navigation

### Actual behavior
The scan job restarts scanning every time when another activity is opened.
As a result, the app gets the error: "App 'com.example.myapplication' is scanning too frequently" in the log cat and scanning doesn't work for some time.

class MyApplication : Application(), MonitorNotifier {
  override fun onCreate() {
    super.onCreate()
    val beaconManager = BeaconManager.getInstanceForApplication(this)
    beaconManager.addMonitorNotifier(this)
    beaconManager.startMonitoring(region)
  }
...
}

From documentation:
`There are subtle changes with the initial background state when using autobind. If you use autobind methods to start ranging or monitoring from Application.onCreate or any method in its call stack, then the library will **start scanning in background mode**. If you initiate scanning at any other time (from a custom service or from an activity or fragment) then the library will start out in foreground mode provided that the screen is on. Previously, the library always started out in foreground mode unless using RegionBootstrap, in which case it started out in background mode.
`

As I understood correctly, it should perform background scanning if I have starting of monitoring beacons in the Application.onCreate method. and should not restart scanning every time on navigation. Looks like it doen't work. 

### Steps to reproduce this behavior
1) Launch the sample app
2) Close the app, go to settings and grant all the permissions;
3) Open the app again;
3) See the Job started scanning
4) See logs
Running immediate scan job: instance is org.altbeacon.beacon.service.ScanJob@a1bb557
scanJob version 2.20.2 is starting up on the main process
5) Open next activity by button;

### Mobile device model and OS version
Google Pixel 7 Pro, Android 14

### Android Beacon Library version 2.20.6

Sample:
[BeaconsApplication.zip](https://github.com/AltBeacon/android-beacon-library/files/15465248/BeaconsApplication.zip)

IMPORTANT:  This forum is reserved for feature requests or reproducible bugs with the library itself.  If you need help with using the library with your project, please open a new question on StackOverflow.com.

> 088 ScanJob                 com.example.myapplication            I  Using periodicScanJobId from manifest: 208352940
2024-05-28 09:59:02.297 24088-24088 ScanJob                 com.example.myapplication            I  ScanJob Lifecycle START: org.altbeacon.beacon.service.ScanJob@ecc2771
[...]

PR #1195, opened 2024-05-29
Fixed issue when scan job could restart every time when we have activity navigation

VolodaUa:
The fix fixes the issue when we have the scan job restarting every time during navigation (activity opening) in the app.

Background mode - is set only if the user puts the app in the background or it keeps the background when startMonitoring was called in Application.onCreate during app start(as designed for background scanning).
Foreground mode - is set only one time when app become foreground. Fixed also an issue when starting activity is like the facade to find the proper starting activity to proceed app starting.

Fixes: https://github.com/AltBeacon/android-beacon-library/issues/1194

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -79,6 +79,8 @@ dependencies {
     androidTestImplementation 'org.apache.commons:commons-math3:3.6.1'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
     implementation "androidx.core:core-ktx:1.3.2" // 1.10.1 requires targeting android SDK 33
+    implementation "androidx.lifecycle:lifecycle-process:2.6.2"
+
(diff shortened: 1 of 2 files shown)


VolodaUa:
@davidgyoung  Please take a look 

Thanks a lot!

VolodaUa:
I am thinking about using ProcessLifecycleOwner.get().getLifecycle() as it's more accurate that activities..

VolodaUa:
Hi @davidgyoung 

Could you please prepare a beta version with this fix? I locally tested it a lot and it works.

VolodaUa:
Hi @davidgyoung 

We tested this solution with QA for a while. The application faced this issue once we started using this library. The fix solved the problem and then the app worked well. No more complains. I had to use reflection to replace the affected class in the library with my own. It's not good, but it the solution for now.  
The application can't detect beacons after simple navigation.

Please let me know if I need to provide more information.

davidgyoung:
Thank you @VolodaUa

PR #1229, opened 2025-03-09
warn user about misconfigured manifest for beacon scanning

davidgyoung:
This is designed to help users who can't detect beacons because they don't have permissions declared in their manifest, especially including the neverForLocation permission attribute that may be inserted by other third party libraries.


lib/src/main/java/org/altbeacon/beacon/utils/PermissionsInspector.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        if (!hasPermission(android.Manifest.permission.ACCESS_BACKGROUND_LOCATION)) {
+            LogManager.w(
+                TAG,
+                "ACCESS_BACKGROUND_LOCATION permission not in manifest.  Will not be able to scan for bluetooth beacons"
+            )
+        }
+
+        return ok
+    }
```
echedey-ls, line 50: Won't this function always return false?

davidgyoung: You are correct, that you.  Need to change to set var ok = false

Issue #1221, opened 2025-02-12
2.21.0-beta5 should respect the background periods properly


elican-doenyas:
Currently, when the library user configures scan periods for background scanning, the library discards the backgroundBetweenScanPeriod, and uses the backgroundScanPeriod for that instead. This is present in 2 places in BeaconManager class, both in the applySettingsChange method.

1.  Line 258

![Image](https://github.com/user-attachments/assets/74e9b718-9e46-49ab-86be-3c5033a8639a)

2. Line 275

![Image](https://github.com/user-attachments/assets/b98b781f-dc09-44ac-be6a-990b5eb51388)

Also, while I am not sure of this, the default value for backgroundBetweenScanPeriod in Settings.kt file, ScanPeriods data class should not be 0.

@davidgyoung  thanks in advance.

elican-doenyas:
@davidgyoung any update on this issue? Anything I can do to assist on the resolution of the matter?

davidgyoung:
Thanks for this report, @elican-doenyas.  This is a bug and needs to be fixed.  Would you like to submit a Pull Request with this change?  I will gladly merge.

We can also change the default value for the backgroundBetweenScanPeriod in Settings.kt to be 300000 ms (5 minutes) which matches the default elsewhere in the library.

davidgyoung:
Fixed in#1221

PR #1230, opened 2025-03-10
ISSUE#1221: Use correct scanning parameters for bg scan

elican-doenyas:
Fixes the issue reported here: https://github.com/AltBeacon/android-beacon-library/issues/1221


davidgyoung (approved):
Looks good, thank you!

